PR #1, opened 2020-08-09
Fix many styling issues

fushinari:
Feel free to close the PR if you don't like the changes or you don't want to follow the [PEP8 style guide](https://www.python.org/dev/peps/pep-0008). This PR doesn't solve all PEP8 issues anyway. Mainly, [the naming conventions](https://www.python.org/dev/peps/pep-0008/#naming-conventions), docstrings, line length limit etc. But as is, I think the code is pretty readable.

I've also removed unused imports and import LANGUAGES from googletrans instead of importing googletrans as a whole.

If you do merge, please squash and merge because I only made it multiple commits to make the PR itself more easy to review.

Thanks

### DIFF
--- a/gnabel.py
+++ b/gnabel.py
@@ -327,6 +327,7 @@ class MainWindow(Gtk.Window):
         AboutText.set_website("https://github.com/gi-lom/gnabel")
         AboutText.set_website_label("Github page")
         AboutText.set_logo(None)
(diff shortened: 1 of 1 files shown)


fushinari:
I've also fixed the "Copy to Clipboard" function.

gi-lom:
Looks good, now the code will be even more readable. And thank you for solving the clipboard issue!

PR #5, opened 2020-08-09
Use standard paths

KSXGitHub:
Fixes https://github.com/gi-lom/gnabel/issues/2
Fixes https://github.com/gi-lom/gnabel/issues/3

## Changes

1. Executable script (`gnabel.py`) is now copied directly to bin directory (`/usr/bin/gnabel` or `~/.local/bin/gnabel`). Entering `gnabel` in terminal would launch the application. Desktop file can now use `Exec=gnabel` to define execution command.
2. Icon is now installed using `xdg-icon-resource install --size 64 icon.png gnabel` command. Desktop file can now use `Icon=gnabel` to define icon.
3. Content of would-be desktop file is now an actual file (`gnabel.desktop`). It can now be copied without modification.
4. Settings file is now at `~/.config/gnabel/settings.json`.
5. I have also used [shellcheck](https://www.shellcheck.net/) to fix potential bugs in `install.sh`.


install.sh
```diff
@@ -19,28 +26,9 @@ echo "Installing..."
...
-Type=Application
-Terminal=false
-Exec=python3 ${DEST_DIR}/gnabel/gnabel.py
-Name=Gnabel
-Comment=A translation app for GTK environments based on Google Translate.
-Icon=${DEST_DIR}/gnabel/icon.png
-Keywords=gnabel;translate;translation
-EOT
-mv gnabel.desktop ${DEST_DIR}/applications
+cp -v gnabel.py "$DEST_DIR/gnabel.py"
```
fushinari, line 29: It's better to do `cp -v gnabel.py "$DEST_DIR/gnabel"` instead. It's not nice to have files in bin folders with file extensions.

install.sh
```diff
@@ -1,13 +1,20 @@
...
 
 # Destination directory
 if [ "$UID" -eq "$ROOT_UID" ]; then
-  DEST_DIR="/usr/share/"
+  DEST_DIR="/usr/bin"
+  ICON_DIR="/usr/share/default"
 else
-  DEST_DIR="$HOME/.local/share"
+  DEST_DIR="$HOME/bin"
+  ICON_DIR="$HOME/.icons"
```
fushinari, line 17: `$HOME/.icons` shouldn't be a thing anymore. Use `$HOME/.local/share/icons`.

Ideally, it would check for the `$XDG_DATA_HOME` environment variable and use `$XDG_DATA_HOME/icons` and if `$XDG_DATA_HOME` is not defined, fallback to `$HOME/.local/share/icons`.

KSXGitHub:
@fushinari I have complied with your requests. Do you have any other criticisms?

fushinari:
> @fushinari I have complied with your requests. Do you have any other criticisms?

Nope, looks great

fushinari:
Sorry, I missed something. The icon won't work.

It needs to follow Icon Theme Specification. I think this would be the necessary read: https://specifications.freedesktop.org/icon-theme-spec/icon-theme-spec-latest.html#directory_layout

install.sh
```diff
@@ -19,28 +32,9 @@ echo "Installing..."
...
-Terminal=false
-Exec=python3 ${DEST_DIR}/gnabel/gnabel.py
-Name=Gnabel
-Comment=A translation app for GTK environments based on Google Translate.
-Icon=${DEST_DIR}/gnabel/icon.png
-Keywords=gnabel;translate;translation
-EOT
-mv gnabel.desktop ${DEST_DIR}/applications
+cp -v gnabel.py "$BIN_DIR/gnabel"
+cp -v icon.png "$ICON_DIR/gnabel.png"
```
fushinari, line 36: Use `xdg-icon-resource install --size 64 icon.png gnabel`

KSXGitHub:
@fushinari I have switched to `xdg-icon-resource install --size 64 icon.png gnabel`. I also added some other changes, can you review it for me?

gnabel.py
```diff
@@ -17,6 +17,7 @@
 LanNumber = 8 #number of language tuples to save in history
 ButtonLength = 65 #length of language buttons
 ButtonNumLanguages = 3 #number of language buttons
+SettingsFile = os.path.expanduser('~/.config/gnabel/settings.json')
```
fushinari, line 20: All references to `~/.config` should be using `XDG_CONFIG_HOME` with `~/.config` as fallback if it's not set.

GLib is already being imported (unnecessarily until now), so you can just use `GLib.get_user_data_dir()` for getting the necessary directory.

gi-lom:
Thank you for your time guys!

gnabel.py
KSXGitHub: I will assume that `GLib.get_user_data_dir()` takes no arguments and always return a string.

fushinari: Yes, nvm though. He already merged this.

KSXGitHub: I can make another PR

PR #10, opened 2020-08-10
Initial Flatpak manifest

rafaelmardojai:
If you're interested in distributing **gnabel** as a **Flatpak**, here's an initial manifest.

Python dependencies are splitted into `pypi-dependencies.json` file.


gi-lom:
Looks good. Can I ask if you generated these scripts through an automated tool?

rafaelmardojai:
Main manifest was manually made. But `pypi-dependencies.json` was generated using **pip** tool from https://github.com/flatpak/flatpak-builder-tools

PR #19, opened 2020-08-12
Unify install.sh and uninstall.sh into a single script

KSXGitHub:
## Before

- References (to files and libraries) are duplicated, prone to errors should changes would be made.
- Uninstall script is already buggy, it must be fixed anyway.
- ~~Install script uses `sudo` for `pip3`, which is unnecessary for root user and incorrect for non-root user.~~

## After

- No duplicated references. Everything is in `install.sh`.
- Correctly use `--user` for `pip3` for non-root user.


KSXGitHub:
@gi-lom I have noticed in your recent change that you wish (intentionally or unintentionally) to add `--user` to `pip3` regardless of whether the install/uninstall script is run with root privileged or not. I have resolve merge conflict by deciding that `--user` won't be added on root, but please tell me if this is against your wish.

KSXGitHub:
@fushinari Review?

fushinari:
The meson build system in #20 works well. So, I think this is not necessary anymore. It was great though.

gi-lom:
Since installing scripts are no more necessary, I will close this

KSXGitHub:
You should delete `install.sh` and `uninstall.sh` to avoid confusion.

PR #20, opened 2020-08-12
Use meson build system

rafaelmardojai:
### Meson
- Rename `icon.svg` to `gnabel.svg`
- Move `gnabel.desktop` and `gnabel.svg` to `data` folder
- Move `gnabel.py` to `src` folder
- Add `meson.build` files to root, `data` and `src` folders
- Add `postinstall.py` on `build-aux/meson`

### Flatpak
- Move flatpak related files to `build-aux/flatpak` folder
- Use meson build on flatpak manifest 

### TODO
- [x] Add build instructions on README
- [ ] ~~Add localization support~~


rafaelmardojai:
Cc @gi-lom @fushinari @KSXGitHub

rafaelmardojai:
About the app ID, there's the convention to use the [reverse domain format](https://en.wikipedia.org/wiki/Reverse_domain_name_notation). In our case it would be `com.github.gi_lom.gnabel`.

I can rename the `.desktop` file and icon to use it, if you are fine with it. We would also use the same ID for `Gtk.Application` and `GResource` if we use them later.

fushinari:
Looks great. I like it.

rafaelmardojai:
I just rebased. Localization and app ID change can be done later.
@gi-lom feel free to merge.

Issue #25, opened 2020-08-13
HIG compliant app icon


jimmac:
While Gnabel ships an app icon, it would be nice to have a proper [GNOME HIG compliant](https://developer.gnome.org/hig/stable/icon-design.html.en) app icon with a symbolic variant.

jimmac:
Might show its origins a little strongly, but here's my initial suggestion:

![image](https://user-images.githubusercontent.com/14821/90127283-39d64300-dd65-11ea-917e-127f7e559d63.png)

gi-lom:
Made a symbolic icon

PR #30, opened 2020-08-15
Many improvements, fixes, etc

fushinari:
### Changes
- ButtonBoxes where appropriate
- Better widget positioning in HeaderBar
- About dialog is now a proper modal window
- Made Gnabel a proper GtkApplication
- PopoverMenu to replace current About button and follow GNOME HIG
- Fix icons again (after app id change)
- Remove deprecated VBoxes and HBoxes, replaced by Boxes
- Fix translation button sensitivity issue
- Add description in about window

Some of the code in here are also in preparation for the future. For example, using GtkApplication, the PopoverMenu, etc. It's for the GNOME HIG and also for ease of future maintenance.

The screenshot has also been updated to match how it would look like on a vanilla GNOME setup (except the font I think?).

preview

This would've been even easier if Gnabel used Glade and if I didn't focus on keeping it a single-file script but well, this works and it works pretty well as far as I can see.

### Bigger questions

@rafaelmardojai @KSXGitHub @gi-lom 

I'm pinging you all because I want to know your opinions.

I'd also like to know your thoughts about these things:
- Switching to Glade for the GUI
- Splitting the code into multiple files and using gresource
- More keyboard accessibility and Shortcuts window
- libhandy (for mobile interface)

**Most importantly, are we really following GNOME HIG?**

The current icon doesn't follow the GNOME HIG (#25) and the UI didn't until now (and even now, has some issues remaining). 
So, is Gnabel a GNOME application or is it just an application that works best in GNOME?


rafaelmardojai:
> Switching to Glade for the GUI
> Splitting the code into multiple files and using gresource

That will be very welcome to make source code cleaner.

> More keyboard accessibility and Shortcuts window

+1

> libhandy (for mobile interface)

Also take advantage to make some mockups for a more HIG compliant UI.

### About PR
- Right side headerbar buttons shouldn't be linked. They're not related to each other.
- The "Quit" menu entry isn't any more used by GNOME apps.
- I think you can remove `self.set_icon_name("com.github.gi-lom.gnabel")` since `GtkApplication` will take care of that.

src/gnabel.py
```diff
@@ -61,15 +81,16 @@ class MainWindow(Gtk.ApplicationWindow):
                 json.dump(Settings, outfile, indent=2)
 
     # Mount everything
-    def __init__(self):
+    def __init__(self, app):
         self.Translator = Translator()
-        Gtk.Window.__init__(self, title="GTranslate")
+        Gtk.Window.__init__(self, title="Gnabel", application=app)
```
rafaelmardojai, line 86: This should be `Gtk.ApplicationWindow`

fushinari:
> Also take advantage to make some mockups for a more HIG compliant UI.

Sorry, I didn't understand you.

> Right side headerbar buttons shouldn't be linked. They're not related to each other.

Yes, I agree. I'll revert that to just a Box instead of a ButtonBox.

> The "Quit" menu entry isn't any more used by GNOME apps.

I didn't know until now. Thanks. But there's an issue with removing this. We want the PopoverMenu because "About Gnabel" (i agree with the naming of the menu item btw) is supposed to be in the menu according to the GNOME HIG. But then there's nothing else in the menu except "About Gnabel". What do you think about moving the clipboard option to the menu?

> I think you can remove `self.set_icon_name("com.github.gi-lom.gnabel")` since `GtkApplication` will take care of that.

I've decided to use `set_default_icon_name` instead because that's what GNOME Music does.

rafaelmardojai:
> Sorry, I didn't understand you.

I wanted to say that we should work in some redesign mockups, and take libhandy in count.

> I didn't know until now. Thanks. But there's an issue with removing this. We want the PopoverMenu because "About Gnabel" (i agree with the naming of the menu item btw) is supposed to be in the menu according to the GNOME HIG. But then there's nothing else in the menu except "About Gnabel". What do you think about moving the clipboard option to the menu?

We can keep it alone for now. We will need a settings window soon.

> I've decided to use set_default_icon_name instead because that's what GNOME Music does.

Yeah, you can keep it as fallback. But any decent DE will be able to link the app ID with the desktop file.

fushinari:
> I wanted to say that we should work in some redesign mockups, and take libhandy in count.

Ah, I agree. The current UI is not good when it comes to mobile interfaces.

> We can keep it alone for now. We will need a settings window soon.

Okay, gonna keep it alone then.

> Yeah, you can keep it as fallback. But any decent DE will be able to link the app ID with the desktop file.

True.

fushinari:
@rafaelmardojai Is it okay now?

rafaelmardojai:
@fushinari LGTM!

rafaelmardojai:
Could you also add the symbolic icon to the meson build? It goes to the same location `icons/hicolor/scalable/apps`.

fushinari:
@rafaelmardojai Done :)

gi-lom:
>     * Switching to Glade for the GUI
How much work would the trnasition cost? Is the new code designed for that?
>     * Splitting the code into multiple files and using gresource
Yes
>     * More keyboard accessibility and Shortcuts window
Yes, we need a discussion about what shortcuts to add
>     * libhandy (for mobile interface)
Yes, making it mobile friendly was one of the first requests when I first showed the app on Reddit
> 
> **Most importantly, are we really following GNOME HIG?**
> 
> The current icon doesn't follow the GNOME HIG (#25) and the UI didn't until now (and even now, has some issues remaining).
> So, is Gnabel a GNOME application or is it just an application that works best in GNOME?
The one thing that keeps me from saying it could be a full GNOME application is the fact it uses Google translate

fushinari:
> How much work would the trnasition cost? Is the new code designed for that?

It's relatively easy and yes, some of the new code is focused on making the transition easier. It was already pretty well-adapted to using Glade and now, the rest is also pretty much done. Now, we only need to make a glade file and use it as the template.

To see an example of how the code would look, see [this](https://github.com/fushinari/reo/blob/master/reo/gtk/window.py). It's my dictionary application's GTK interface. As you'll notice immediately, most of the needed bits are already there in Gnabel too.

> Yes, we need a discussion about what shortcuts to add

Ctrl+Enter is already a shortcut for translate. I also suggest Ctrl+Shift+C for copy to clipboard.

> Yes, making it mobile friendly was one of the first requests when I first showed the app on Reddit

Yes, I remember. We need to open an issue for this and create mockups.

> The one thing that keeps me from saying it could be a full GNOME application is the fact it uses Google translate

Why? GNOME has no requirements for GNOME apps to be based only on fully free services. The only real blockers are the UI and the icon (there's a symbolic icon too now, but the icons are not GNOME HIG-compliant). Btw, when I say GNOME Application, I only mean a "Made for GNOME" application. Being a core app is not necessary and is impossible anyway.

rafaelmardojai:
Transition to Glade could be done in one shot with a HIG/libhandy redesign.

gi-lom:
Why the shortcut to translate is now Ctrl+Enter? In most apps just pressing enter executes the command

fushinari:
> Why the shortcut to translate is now Ctrl+Enter? In most apps just pressing enter executes the command

@gi-lom I did that to keep the behavior consistent across all GNOME applications. When we add a preferences window (along with the libhandy redesign), we can add an option to switch this behavior, giving two options:

1. Ctrl+Enter for newline and Enter for translate.
2. Enter for newline and Ctrl+Enter for translate (current behavior).

If you think an alternative behavior is better, please suggest it.

> Transition to Glade could be done in one shot with a HIG/libhandy redesign.

@rafaelmardojai I agree. Could you open an issue for the libhandy redesign? Any design ideas you have in mind? A mobile-only UI or a desktop-only UI would've been easy but a convergent UI is making my mind go blank. I do have an odd idea of using GtkStacks to present different UIs based on window width though...

PR #35, opened 2020-09-14
Use threading wherever necessary

### LINKED ISSUE #17: TTS audio should be downloaded and played asynchronously
## Problem

Currently, whenever user clicks the "speaker" button, the GUI freezes until audio is downloaded and played. This is a bad user experience.

## Suggestion

Use `async` and `await` to download and play audio.

**Expected behavior:**
* When user first clicks the "speaker" button, the button should change its icon to "stop" (as in "play"/"pause"/"stop") and the application should begin downloading audio file from Google.
* While audio is being downloaded or played, if user clicks the "stop" button, the downloading/playing process should be stopped, and the button should change its shape back to "speaker".
* When audio finishes, the "stop" button should change its icon to back to "speaker".

---

**P.S.** I am inexperience in Python so I can't help you with this, sorry :(

fushinari:
Fixes #17 .

@rafaelmardojai There's a possible issue with this implementation. A minor race condition if the Translate button is clicked continuously. In testing, it has never happened but theoretically it can since I set the variables "ActiveThread" and "WaitingThread" from within the thread. I don't know how to fix it. Any suggestions?

### DIFF
--- a/src/dialect.py
+++ b/src/dialect.py
@@ -50,7 +50,6 @@ class MainWindow(Gtk.ApplicationWindow):
     # Language values
     LangCode = list(LANGUAGES.keys())
     LangName = list(LANGUAGES.values())
-    Translator = Translator()
     # Current input Text
     CurrentInputText = ""
     CurrentHistory = 0
@@ -151,9 +150,9 @@ class MainWindow(Gtk.ApplicationWindow):
         self.FirstLanguageCombo.connect("changed", self.HistoryLeftLanUpdate)
 
         ### Switch
-        Switch = Gtk.Button.new_from_icon_name("object-flip-horizontal-symbolic", Gtk.IconSize.BUTTON)
-        Switch.set_tooltip_text("Switch languages")
-        Switch.connect("clicked", self.UISwitch)
+        self.Switch = Gtk.Button.new_from_icon_name("object-flip-horizontal-symbolic", Gtk.IconSize.BUTTON)
+        self.Switch.set_tooltip_text("Switch languages")
+        self.Switch.connect("clicked", self.UISwitch)
 
         ### Second language
         SecondLanguageList = Gtk.ListStore(str)
@@ -171,7 +170,7 @@ class MainWindow(Gtk.ApplicationWindow):
         LanguageButtonBox.set_layout(Gtk.ButtonBoxStyle.EXPAND)
         LanguageButtonBox.set_homogeneous(False)
         LanguageButtonBox.pack_start(self.FirstLanguageCombo, True, True, 0)
-        LanguageButtonBox.pack_start(Switch, True, False, 0)
+        LanguageButtonBox.pack_start(self.Switch, True, False, 0)
         LanguageButtonBox.pack_start(self.SecondLanguageCombo, True, True, 0)
 
         ### Voice
@@ -310,8 +309,31 @@ class MainWindow(Gtk.ApplicationWindow):
         ll = button.get_label()
         self.SecondLanguageCombo.set_active(self.LangName.index(ll.lower()))
 
+    def SwitchAll(self, FirstLanguage, SecondLanguage, FirstText, SecondText):
+        self.FirstLanguageCombo.set_active(self.LangCode.index(SecondLanguage) + 1)
+        self.SecondLanguageCombo.set_active(self.LangCode.index(FirstLanguage))
+        self.LeftBuffer.set_text(SecondText)
+        self.RightBuffer.set_text(FirstText)
+
+        # Re-enable widgets
+        self.FirstLanguageCombo.set_sensitive(True)
+        self.SecondLanguageCombo.set_sensitive(True)
+        self.Switch.set_sensitive(True)
+
+    def SwitchAutoLang(self, SecondLanguagePos, FirstText, SecondText):
+        RevealedLanguage = str(self.Translator.detect(FirstText).lang)
+        FirstLanguagePos = self.LangCode.index(RevealedLanguage) + 1
+        FirstLanguage = self.LangCode[FirstLanguagePos - 1]
+        SecondLanguage = self.LangCode[SecondLanguagePos]
+
+        # Switch all
+        GLib.idle_add(self.SwitchAll, FirstLanguage, SecondLanguage, FirstText, SecondText)
+
     def UISwitch(self, button):
         # Get variables
+        self.FirstLanguageCombo.set_sensitive(False)
+        self.SecondLanguageCombo.set_sensitive(False)
+        self.Switch.set_sensitive(False)
         FirstBuffer = self.LeftBuffer
         SecondBuffer = self.RightBuffer
         FirstLanguagePos = self.FirstLanguageCombo.get_active()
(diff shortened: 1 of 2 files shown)


rafaelmardojai:
Mmm, there isn't conditional code to check if the text to translate has changed at all? Because if the text is the same the Translate button should do nothing.

src/dialect.py
```diff
@@ -488,42 +500,61 @@ def Translation(self, button):
...
+                self.Return.set_sensitive(True)
+            if len(self.Settings["Translations"]) == TransNumber:
+                self.Settings["Translations"].pop()
+            self.Settings["Translations"].insert(0, NewHistoryTrans)
+            # Save everything in the JSON file
+            with open(SettingsFile, 'w') as outfile:
+                json.dump(self.Settings, outfile, indent=2)
+
+        self.ActiveThread = self.WaitingThread
+        self.WaitingThread = None
```
fushinari, line 553: @rafaelmardojai This is the problem. I'm switching the active thread here. But the issue is that if the user clicks the Translate button while this is happening, L504 to L515 might get executed at conflicting times. I don't know what will be the end result but it will be bad.

rafaelmardojai: Yeah, now i understand what you where saying. 

What about using the `queue` python module? I have never used it, but maybe will solve our issue.

rafaelmardojai:
What if we have a permanent ruining thread for the translation?

We have a list called `self.queue` where we store the strings to translate every.

When the user clicks the Translate button we add the text to the queue with something like:

```python
def add_to_queue(self):
    text =  # get self.LeftBuffer text
    # apply any conditional to the text
    old = self.queue.pop(0) # remove previous text from queue
    self.queue.append(text) # add text to queue
```
The permanent thread could look like:

```python
def Translation(self):
    while True:
        if self.queue:
            text = self.queue.pop(0)
            # Translate and do wherever with the text 
```

fushinari:
> What if we have a permanent ruining thread for the translation?
> 
> We have a list called `self.queue` where we store the strings to translate.
> 
> When the user clicks the Translate button we add the text to the queue with something like:
> 
> ```python
> def add_to_queue(self):
>     text =  # get self.LeftBuffer text
>     # apply any conditional to the text
>     old = self.queue.pop(0) # remove previous text from queue
>     self.queue.append(text) # add text to queue
> ```
> 
> The permanent thread could look like:
> 
> ```python
> def Translation(self):
>     while True:
>         if self.queue:
>             text = self.queue.pop(0)
>             # Translate and do wherever with the text 
> ```

Nice, this could work. Let me try.

fushinari:
@rafaelmardojai Can you test the current implementation? It's working well on my end. It doesn't do the cleaning of unnecessary requests anymore, but then again, maybe that was a bad idea anyway because it goes against how threads work.

fushinari:
Re-added the cleaning and just hope that the GIL means what I think it means, lol.

fushinari:
Is it just me or is gTTS not working anymore?

rafaelmardojai:
> Is it just me or is gTTS not working anymore?

Yeah, see: https://github.com/pndurette/gTTS/issues/226

rafaelmardojai:
> @rafaelmardojai Can you test the current implementation?

All looks good and smooth on my end.

We need to add some feedback to let the user know that something is happening when he click the Translate button. 
Making the `LeftText` insensitive? Adding a spinner somewhere?

fushinari:
> > @rafaelmardojai Can you test the current implementation?
> 
> All looks good and smooth on my end.
> 
> We need to add some feedback to let the user know that something is happening when he click the Translate button.
> Making the `LeftText` insensitive? Adding a spinner somewhere?

We could do the same thing as the audio button. Putting a spinner on the button I mean. But then, that could make the user assume that he/she can't click "Translate" again until it's done. I don't see any other logical position for the spinner to be in either...

rafaelmardojai:
Other issue is that the app takes a while to show up, probably because slow internet connection. Do we make a request to the API on startup?

fushinari:
> Other issue is that the app takes a while to show up, probably because slow internet connection. Do we make a request to the API on startup?

Yes, for language codes. The app will not even launch without internet. I want to change this, but a lot of the code depends on it.

rafaelmardojai:
> We could do the same thing as the audio button. Putting a spinner on the button I mean. But then, that could make the user assume that he/she can't click "Translate" again until it's done. I don't see any other logical position for the spinner to be in either...

Could we change the left text view with a spinner when translating? Or add some points (...) at the end of the left text and make the view insensitive as Google Translate.

rafaelmardojai:
> Yes, for language codes. The app will not even launch without internet. I want to change this, but a lot of the code depends on it.

Right, we should show a empty window with a spinner on startup and also have some error view.

fushinari:
> Right, we should show a empty window with a spinner on startup and also have some error view.

I did add an error dialog when no network is available. But a splash screen also seemed like a bad idea since no other GNOME app does it. Maybe use a stack with a spinner on one page and the actual content on another page and switch pages when loading is done?

fushinari:
> > We could do the same thing as the audio button. Putting a spinner on the button I mean. But then, that could make the user assume that he/she can't click "Translate" again until it's done. I don't see any other logical position for the spinner to be in either...
> 
> Could we change the left text view with a spinner when translating? Or add some points (...) at the end of the left text and make the view insensitive as Google Translate.

Then the ability to type while the current translation is in progress goes away. Or should we only keep that ability while in live mode?

rafaelmardojai:
> Then the ability to type while the current translation is in progress goes away. Or should we only keep that ability while in live mode?

LOL, Sorry, I wanted to say the RIGHT text. 😅

rafaelmardojai:
> I did add an error dialog when no network is available. But a splash screen also seemed like a bad idea since no other GNOME app does it. Maybe use a stack with a spinner on one page and the actual content on another page and switch pages when loading is done?
 
I was thinking something like my [Webfont Kit Generator](https://github.com/rafaelmardojai/WebfontKitGenerator) app progress view:

fushinari:
> LOL, Sorry, I wanted to say the RIGHT text. :sweat_smile:

Lol, it's fine. Adding a "..." to the end is easy, like Google Translate, but we might want to put off adding proper feedback for later. I mean, at this point, it's a given that we need to re-do the whole UI in Glade. Might be better to wait until then for the loading view too.

rafaelmardojai:
Yeah. I think this PR is ready to merge.

When merged I will work on an initial split of files.

fushinari:
> 
> I was thinking something like my [Webfont Kit Generator](https://github.com/rafaelmardojai/WebfontKitGenerator) app progress view:
> 
> So we should have three views:
> 
>   * Loading
>   * Network error
>   * Translator

Yep, this was almost what I had in mind too. But adding more UI elements in this code would be a waste of time. I can only do proper Handy-based development when GNOME 3.38 is out because my environment is a mess right now (libhandy and libhandy1 conflicts in Glade and such). That's my fault entirely but I haven't had the time to fix it. (The AUR `libhandy1` package is also mine and I haven't looked into all the issues it has... People are using it though... They don't know what I'm putting them through :sweat_smile: )

fushinari:
@gi-lom It's okay to merge now.

rafaelmardojai:
Just noticed that switching languages will freeze the UI, not too much noticeable, so we can fix it later.

fushinari:
> Just noticed that switching languages will freeze the UI, not too much noticeable, so we can fix it later.

Yes, there's a call to the translator there too. L325, Should I thread it?

rafaelmardojai:
> Should I thread it?

Ideally yes. It takes a while when the language is in Auto.

rafaelmardojai:
Also noticed `self.Translator` is defined twice, L53 & L86.

fushinari:
> Ideally yes. It takes a while when the language is in Auto.

I've fixed this and corrected the double definition. I'm noticing that the current history storage mechanism is unreliable. But that's for later.

rafaelmardojai:
> I've fixed this and corrected the double definition. I'm noticing that the current history storage mechanism is unreliable. But that's for later.

Yeah, I'm also suggesting using GSettings (#36)

rafaelmardojai:
@gi-lom merge?

gi-lom:
Sorry for the delay

PR #37, opened 2020-09-24
Initial files split and gresource

rafaelmardojai:
What does this PR?

- Implements the meson setup needed to split source files
  - Generates the bin file from `dialect.in`.
  - Install all the sources listed in the meson file (right now only `main.py`).
  - The bin file imports `main.py` and run the app
- Implements the code to start using GResource. Currently we only have the `menu.ui` resource for the app menu.
- Also added license/copyright notice to every python file.

Any other change I can do in this PR?

cc @fushinari

### DIFF
--- a/data/meson.build
+++ b/data/meson.build
@@ -1,3 +1,5 @@
+subdir('resources')
+
 # Install desktop
 install_data(
   'com.github.gi_lom.dialect.desktop',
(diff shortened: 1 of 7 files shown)


fushinari:
LGTM.

rafaelmardojai:
> LGTM! Maybe we should use SPDX headers too?

Switched to SPDX headers!

PR #39, opened 2020-09-30
Code cleanup

fushinari:
- PEP8: snake_case naming style.
- PEP8: 120 characters limit
- PEP8: Reorder imports
- Rename some variables to make more sense
- Fix handling of connectivity issues on launch

I think it works fine. Just wanted it to get tested a little before being merged.

### DIFF
--- a/meson.build
+++ b/meson.build
@@ -5,11 +5,11 @@ project('dialect',
                    ],
 )
 
-python = import('python3')
+python = import('python')
 
 message('Looking for dependencies')
-python_bin = python.find_python()
-if not python_bin.found()
+py_installation = python.find_installation('python3')
+if not py_installation.found()
     error('No valid python3 binary found')
 else
     message('Found python3 binary')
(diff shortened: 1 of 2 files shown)


rafaelmardojai:
Awesome! 

I have no time to test it, but checking the code all looks good.

Should we also use a variable for the GResource path, like we do with the app ID?

rafaelmardojai (approved):

fushinari:
> Should we also use a variable for the GResource path, like we do with the app ID?

GResource path? You mean the UI file?

rafaelmardojai:
> GResource path? You mean the UI file?

Yeah, I mean, the UI files have a GResource path similar to the app ID. 

If we are going to have more UI files, probably makes sense to have the base path in a variable.

fushinari:
I've added a global variable RES_PATH for the purpose. If it looks okay, please merge. :)

PR #43, opened 2020-10-05
Rewrite

### LINKED ISSUE #32: libhandy UI
We need mockups and ideas for implementation of a libhandy-based Mobile user interface. It has to be usable both on mobile and desktop and should not feel out of place in either environments.

### LINKED ISSUE #40: Add preferences window
What preferences will we want to expose? 

Some thoughts:

- Option to enable automatic translation as you type (with some warning about API abuse)
- Option to disable translation history?
- Button to clear translation history?

rafaelmardojai:
**Actual changes:**

- Split a little bit more the source (main.py, define.py and window.py)
- Add libhandy dependency
- Use Gtk.Builder for about dialog
- Add a working preferences window
- Initial window.py rewrite using Gtk.Builder
- Implement loading and error view

**TODO:**
- [x] Rewrite history system
- [x] Fix TTS implementation
- [x] Drop pydub dependency
- [x] Implement de Translate Accel preference on the translation logic
- [x] Implement clear action
- [x] Implement paste action
- [x] Implement final translation progress feedback
- [x] Create a custom widget for the lang selection based on the mockups
- [x] Fix some styling issues

Closes #32 
Closes #40 

I know I'm missing something, so please let me know 😂

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -7,10 +7,11 @@
 import sys
 
 import gi
-gi.require_version('Gtk', '3.0')
 gi.require_version('Gdk', '3.0')
+gi.require_version('Gtk', '3.0')
+gi.require_version('Gst', '1.0')
 gi.require_version('Handy', '1')
-from gi.repository import Gdk, Gio, GLib, Gtk, Handy
+from gi.repository import Gdk, Gio, GLib, Gtk, Gst, Handy
 
 from dialect.define import APP_ID, RES_PATH
 from dialect.window import DialectWindow
@@ -40,8 +41,8 @@ class Dialect(Gtk.Application):
         GLib.set_prgname('com.github.gi_lom.dialect')
         self.setup_actions()
 
-        # Init Handy
-        Handy.init()
+        Handy.init()  # Init Handy
+        Gst.init(None)  # Init Gst
 
         # Load CSS
         css_provider = Gtk.CssProvider()
--- a/src/main.py
+++ b/src/main.py
@@ -12,7 +12,7 @@ import gi
 gi.require_version('Gtk', '3.0')
 gi.require_version('Gdk', '3.0')
 gi.require_version('Handy', '1')
-from gi.repository import Gio, GLib, Gtk, Handy
+from gi.repository import Gio, GLib, Gdk, Gtk, Handy
 
 from dialect.define import APP_ID, RES_PATH
 from dialect.window import DialectWindow
@@ -45,6 +45,13 @@ class Dialect(Gtk.Application):
 
         # Init Handy
         Handy.init()
+
+        # Load CSS
+        css_provider = Gtk.CssProvider()
+        css_provider.load_from_resource(f'{RES_PATH}/style.css')
+        screen = Gdk.Screen.get_default()
+        style_context = Gtk.StyleContext()
+        style_context.add_provider_for_screen(screen, css_provider, Gtk.STYLE_PROVIDER_PRIORITY_USER)
 
     def setup_actions(self):
         """ Setup menu actions """
--- a/src/main.py
+++ b/src/main.py
@@ -14,9 +14,9 @@ gi.require_version('Gdk', '3.0')
 gi.require_version('Handy', '1')
 from gi.repository import Gio, GLib, Gtk, Handy
 
-from .define import APP_ID, RES_PATH
-from .window import DialectWindow
-from .preferences import DialectPreferencesWindow
+from dialect.define import APP_ID, RES_PATH
+from dialect.window import DialectWindow
+from dialect.preferences import DialectPreferencesWindow
 
 
 class Dialect(Gtk.Application):
--- a/src/main.py
+++ b/src/main.py
@@ -12,7 +12,7 @@ import gi
 gi.require_version('Gtk', '3.0')
 gi.require_version('Gdk', '3.0')
 gi.require_version('Handy', '1')
-from gi.repository import Gio, GLib, Gdk, Gtk, Handy
+from gi.repository import Gdk, Gio, GLib, Gtk, Handy
 
 from dialect.define import APP_ID, RES_PATH
 from dialect.window import DialectWindow
--- a/src/main.py
+++ b/src/main.py
@@ -32,7 +32,6 @@ class Dialect(Gtk.Application):
                 application=self,
                 title='Dialect'
             )
-        self.window.show_all()
         self.window.present()
 
     def do_startup(self):
--- a/src/main.py
+++ b/src/main.py
@@ -62,13 +62,13 @@ class Dialect(Gtk.Application):
         about_action.connect('activate', self.on_about)
         self.add_action(about_action)
 
-    def on_preferences(self, action, param):
+    def on_preferences(self, _action, _param):
         """ Show preferences window """
         window = DialectPreferencesWindow()
         window.set_transient_for(self.window)
         window.present()
 
-    def on_about(self, action, param):
+    def on_about(self, _action, _param):
         """ Show about dialog """
         builder = Gtk.Builder.new_from_resource(f'{RES_PATH}/about.ui')
         about = builder.get_object('about')
--- a/src/main.py
+++ b/src/main.py
@@ -16,6 +16,7 @@ from gi.repository import Gio, GLib, Gtk, Handy
 
 from .define import APP_ID, RES_PATH
 from .window import DialectWindow
+from .preferences import DialectPreferencesWindow
 
 
 class Dialect(Gtk.Application):
@@ -46,12 +47,24 @@ class Dialect(Gtk.Application):
         Handy.init()
 
     def setup_actions(self):
-        """Setup menu actions."""
+        """ Setup menu actions """
+
+        preferences_action = Gio.SimpleAction.new('preferences', None)
+        preferences_action.connect('activate', self.on_preferences)
+        self.add_action(preferences_action)
+
         about_action = Gio.SimpleAction.new('about', None)
         about_action.connect('activate', self.on_about)
         self.add_action(about_action)
 
+    def on_preferences(self, action, param):
+        """ Show preferences window """
+        window = DialectPreferencesWindow()
+        window.set_transient_for(self.window)
+        window.present()
+
     def on_about(self, action, param):
+        """ Show about dialog """
         builder = Gtk.Builder.new_from_resource(f'{RES_PATH}/about.ui')
         about = builder.get_object('about')
         about.set_transient_for(self.window)
--- a/src/main.py
+++ b/src/main.py
@@ -5,8 +5,6 @@
 
 # Initial setup
 import sys
-import threading
-from io import BytesIO
 
 import gi
 gi.require_version('Gtk', '3.0')
--- a/src/window.py
+++ b/src/window.py
@@ -66,12 +66,13 @@ class DialectWindow(Handy.ApplicationWindow):
     history = []
     type_time = 0
     trans_queue = []
-    trans_failed = False
     active_thread = None
     # These are for being able to go backspace
     first_key = 0
     second_key = 0
     mobile_mode = False
+    # Connectivity issues monitoring
+    trans_failed = False
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -119,20 +120,23 @@ class DialectWindow(Handy.ApplicationWindow):
         # Get languages available for speech
         threading.Thread(target=self.load_lang_speech).start()
 
-    def load_lang_speech(self):
+    def on_listen_failed(self):
+        self.voice_btn.set_image(self.voice_warning)
+        self.voice_spinner.stop()
+        self.voice_btn.set_tooltip_text('A network issue has occured. Retry?')
+        self.send_notification('A network issue has occured.\nPlease try again.')
+
+    def load_lang_speech(self, search_after=False, second_text=None, second_language_voice=None):
+        """Load the language list for gTTS."""
         try:
             self.lang_speech = list(lang.tts_langs(tld='com').keys())
-            GLib.idle_add(self.toggle_voice_spinner, False)
+            if not search_after:
+                GLib.idle_add(self.toggle_voice_spinner, False)
+            elif second_language_voice in self.lang_speech and second_text != '':
+                self.voice_download(second_text, second_language_voice)
 
         except RuntimeError as exc:
-            def on_fail():
-                self.voice_btn.set_image(self.voice_warning)
-                self.voice_btn.set_sensitive(False)
-                self.voice_spinner.stop()
-                self.voice_btn.set_tooltip_text('No network connection detected.')
-                self.notify('No network connection detected.')
-
-            GLib.idle_add(on_fail)
+            GLib.idle_add(self.on_listen_failed)
             print('Error: ' + str(exc))
 
     def setup_headerbar(self):
@@ -255,7 +259,7 @@ class DialectWindow(Handy.ApplicationWindow):
         self.settings.set_value('right-langs',
                                 GLib.Variant('as', self.right_langs))
 
-    def notify(self, text, timeout=5):
+    def send_notification(self, text, timeout=5):
         """
         Display an in-app notification.
 
@@ -273,7 +277,7 @@ class DialectWindow(Handy.ApplicationWindow):
         )
         timer.start()
 
-    def toggle_voice_spinner(self, active=True, loading=False):
+    def toggle_voice_spinner(self, active=True):
         if active:
             self.voice_btn.set_sensitive(False)
             self.voice_btn.set_image(self.voice_spinner)
@@ -444,10 +448,16 @@ class DialectWindow(Handy.ApplicationWindow):
         # Add here code that changes voice button behavior
         if second_text != '':
             self.toggle_voice_spinner(True)
-            threading.Thread(
-                target=self.voice_download,
-                args=(second_text, second_language_voice)
-            ).start()
+            if self.lang_speech:
+                threading.Thread(
+                    target=self.voice_download,
+                    args=(second_text, second_language_voice)
+                ).start()
+            else:
+                threading.Thread(
+                    target=self.load_lang_speech,
+                    args=(True, second_text, second_language_voice)
+                ).start()
 
     def on_gst_message(self, _bus, message):
         if message.type == Gst.MessageType.EOS:
@@ -464,7 +474,7 @@ class DialectWindow(Handy.ApplicationWindow):
         except Exception as exc:
             print(exc)
             print('Audio download failed.')
-            pass
+            GLib.idle_add(self.on_listen_failed)
         else:
             with NamedTemporaryFile() as file_to_play:
                 tts.write_to_fp(file_to_play)
@@ -472,8 +482,6 @@ class DialectWindow(Handy.ApplicationWindow):
                 self.player.set_property('uri', 'file://' + file_to_play.name)
                 self.player.set_state(Gst.State.PLAYING)
                 self.player_event.wait()
-        finally:
-            # The code to execute no matter what
             GLib.idle_add(self.toggle_voice_spinner, False)
 
     # This starts the translation if Ctrl+Enter button is pressed
@@ -507,7 +515,7 @@ class DialectWindow(Handy.ApplicationWindow):
         self.translate_btn.set_sensitive(sensitive)
         self.clear_btn.set_sensitive(sensitive)
 
-    def user_action_ended(self, buffer):
+    def user_action_ended(self, _buffer):
         if self.settings.get_boolean('live-translation'):
             self.translation(None)
 
@@ -601,14 +609,13 @@ class DialectWindow(Handy.ApplicationWindow):
                     self.trans_failed = False
                 except Exception:
                     self.trans_failed = True
-                    pass
                 GLib.idle_add(self.right_buffer.set_text, second_text)
 
                 # Finally, everything is saved in history
                 self.add_history_entry(first_language, second_language, first_text, second_text)
         if self.trans_failed:
             GLib.idle_add(self.trans_warning.show)
-            GLib.idle_add(self.notify, 'Translation failed.\n Please check for network issues.')
+            GLib.idle_add(self.send_notification, 'Translation failed.\n Please check for network issues.')
             GLib.idle_add(self.copy_btn.set_sensitive, False)
             GLib.idle_add(self.voice_btn.set_sensitive, False)
         else:
--- a/src/window.py
+++ b/src/window.py
@@ -344,6 +344,7 @@ class DialectWindow(Handy.ApplicationWindow):
     def translation(self, _button):
         # If it's like the last translation then it's useless to continue
         if len(self.history) == 0 or not self.appeared_before():
+            self.voice_btn.set_sensitive(True)
             first_buffer = self.left_buffer
             second_buffer = self.right_buffer
             first_text = first_buffer.get_text(first_buffer.get_start_iter(), first_buffer.get_end_iter(), True)
--- a/src/window.py
+++ b/src/window.py
@@ -462,6 +462,7 @@ class DialectWindow(Handy.ApplicationWindow):
 
     def ui_clear(self, _button):
         self.src_buffer.set_text('')
+        self.src_buffer.emit('end-user-action')
 
     def ui_copy(self, _button):
         dest_text = self.dest_buffer.get_text(
--- a/src/window.py
+++ b/src/window.py
@@ -36,6 +36,7 @@ class DialectWindow(Handy.ApplicationWindow):
 
     menu_btn = Gtk.Template.Child()
 
+    char_counter = Gtk.Template.Child()
     src_text = Gtk.Template.Child()
     clear_btn = Gtk.Template.Child()
     paste_btn = Gtk.Template.Child()
@@ -547,7 +548,18 @@ class DialectWindow(Handy.ApplicationWindow):
         self.translate_btn.set_sensitive(sensitive)
         self.clear_btn.set_sensitive(sensitive)
 
-    def user_action_ended(self, _buffer):
+    def user_action_ended(self, buffer):
+        # If the text is over the highest number of characters allowed, it is truncated.
+        # This is done for avoiding exceeding the limit imposed by Google.
+        if buffer.get_char_count() >= MAX_LENGTH:
+            self.send_notification('5000 characters limit reached!')
+            src_text = self.src_buffer.get_text(
+                self.src_buffer.get_start_iter(),
+                self.src_buffer.get_end_iter(),
+                True
+            )
+            self.src_buffer.set_text(src_text[:MAX_LENGTH])
+        self.char_counter.set_text(f'{str(buffer.get_char_count())}/{MAX_LENGTH}')
         if self.settings.get_boolean('live-translation'):
             self.translation(None)
 
@@ -613,7 +625,7 @@ class DialectWindow(Handy.ApplicationWindow):
                     self.trans_spinner.start()
                     self.dest_box.set_sensitive(False)
                     self.langs_button_box.set_sensitive(False)
-                    # If there are not any active threads, create one and start it.
+                    # If there is no active thread, create one and start it.
                     self.active_thread = threading.Thread(target=self.run_translation)
                     self.active_thread.start()
 
@@ -650,10 +662,6 @@ class DialectWindow(Handy.ApplicationWindow):
             # If the two languages are the same, nothing is done
             if src_language != dest_language:
                 dest_text = ''
-                # If the text is over the highest number of characters allowed, it is truncated.
-                # This is done for avoiding exceeding the limit imposed by Google.
-                if len(src_text) > 100:
-                    src_text = src_text[:MAX_LENGTH]
                 # THIS IS WHERE THE TRANSLATION HAPPENS. The try is necessary to circumvent a bug of the used API
                 try:
                     dest_text = self.translator.translate(
@@ -667,7 +675,12 @@ class DialectWindow(Handy.ApplicationWindow):
                 GLib.idle_add(self.dest_buffer.set_text, dest_text)
 
                 # Finally, everything is saved in history
-                self.add_history_entry(src_language, dest_language, src_text, dest_text)
+                self.add_history_entry(
+                    src_language,
+                    dest_language,
+                    src_text,
+                    dest_text
+                )
         if self.trans_failed:
             GLib.idle_add(on_trans_failed)
         else:
--- a/src/window.py
+++ b/src/window.py
@@ -45,6 +45,7 @@ class DialectWindow(Handy.ApplicationWindow):
     translate_btn = Gtk.Template.Child()
 
     right_text = Gtk.Template.Child()
+    trans_spinner = Gtk.Template.Child()
     copy_btn = Gtk.Template.Child()
     voice_btn = Gtk.Template.Child()
 
@@ -102,7 +103,7 @@ class DialectWindow(Handy.ApplicationWindow):
             GLib.idle_add(self.main_stack.set_visible_child_name, 'translate')
 
         except RuntimeError as exc:
-            def quit(self):
+            def quit():
                 sys.exit(1)
 
             self.main_stack.set_visible_child_name('error')
@@ -400,13 +401,7 @@ class DialectWindow(Handy.ApplicationWindow):
                     'first_language': first_language,
                     'second_language': second_language
                 })
-                current_right_text = second_buffer.get_text(
-                    second_buffer.get_start_iter(),
-                    second_buffer.get_end_iter(),
-                    True
-                )
-                if not current_right_text.endswith('...'):
-                    self.right_buffer.set_text(current_right_text + '...')
+                self.trans_spinner.start()
 
                 # Check if there are any active threads.
                 if self.active_thread is None:
@@ -446,6 +441,7 @@ class DialectWindow(Handy.ApplicationWindow):
                     except Exception:
                         pass
                     GLib.idle_add(self.right_buffer.set_text, second_text)
+                    GLib.idle_add(self.trans_spinner.stop)
                     # Finally, everything is saved in history
                     new_history_trans = {
                         'Languages': [first_language, second_language],
--- a/src/window.py
+++ b/src/window.py
@@ -536,8 +536,6 @@ class DialectWindow(Handy.ApplicationWindow):
                 GLib.idle_add(self.left_lang_selector.set_property,
                               'selected', first_language)
                 self.left_langs[0] = first_language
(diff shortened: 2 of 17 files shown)


fushinari (changes requested):
Sorry, I haven't had time to do a proper review. (I guess it's not an issue for now because it's still a draft anyway.) These are just some things I noticed right away.

About the "custom widget" for the language selection, have you seen my glade file on #40 ? It might not be necessary to make a custom widget. Just something that inherits from GtkListBoxRow would be enough to put into two liststores per Combo-Popover. The first liststore for recent languages and the second one for the full languages list. 

Translation progress feedback is what had me thinking the most. I ended up thinking that a spinner on the bottom right of the right textview would be enough but I'm still not really satisfied with it.

rafaelmardojai:
Minor issues, resolved.

Yeah, I was thinking to implement it similar to your glade file. But I was thinking making a "custom widget" will help us to reuse some code.

fushinari:
Some comments:
- Spacing and margin should probably be at least 12 for the first GtkBox.
- The GtkScrolledWindows will also need some styling. (I think you already know that)
- "Reproduce" button doesn't work as it did earlier. Actually, I'm not sure it works at all.
- Should the clear and paste buttons be linked? They're two distinct actions even if they're on the same TextView.
- The buttons under the Left TextView should probably be right-aligned.
- Tooltips for "Clear" and "Paste" buttons

Other things:
- "Copy to Clipboard" can now just be "Copy"
- "Reproduce" should probably be "Listen"

Current:
Screenshot_20201005_183620

After the changes in my comments:
Screenshot_20201005_185329

src/main.py
```diff
@@ -12,7 +12,7 @@
 gi.require_version('Gtk', '3.0')
 gi.require_version('Gdk', '3.0')
 gi.require_version('Handy', '1')
-from gi.repository import Gio, GLib, Gtk, Handy
+from gi.repository import Gio, GLib, Gdk, Gtk, Handy
```
fushinari, line 15: Gdk should be before Gio because alphabetical order.

rafaelmardojai:
Your commend had been adressed.

> "Reproduce" button doesn't work as it did earlier. Actually, I'm not sure it works at all.

It doesn't work if there's no text. I will fix it, so it's insensitive in that case.

rafaelmardojai:
Looking your screenshots, we should ship our own icon for switch language. The semantic of `object-flip-horizontal-symbolic` will broke ours depending on the icons theme.

fushinari:
> Looking your screenshots, we should ship our own icon for switch language. The semantic of `object-flip-horizontal-symbolic` will broke ours depending on the icons theme.

I don't think we should. Since Dialect is a GNOME app, I don't think we should be taking different icon packs into consideration. I'm using Breeze only because I'm on Plasma right now and actually, I think the Breeze icon is very descriptive too. (In fact, I love Breeze's monochrome icons more than anything. Everything else looks unprofessional now. :sweat_smile: Especially on LibreOffice. If I could, I'd use Adwaita with Breeze's monochrome set even on GNOME.)

rafaelmardojai:
I have implemented the new lang selectors. I have the feeling that I have broke something in the process, so please test 😂. There are still some things to polish, but at least it's working.

Now I'm going to sleep. 🤣😂

fushinari:
You didn't sleep??? :laughing:

fushinari:
My exams got over today. Is it okay if I commit directly into this branch or do you want to do this fully?

rafaelmardojai:
Fell free to push. Probably today and tomorrow I wouldn't be able to work on this, so any progress is welcome.

fushinari:
> Fell free to push. Probably today and tomorrow I wouldn't be able to work on this, so any progress is welcome.

Bring it as close as possible to the mockups right?

rafaelmardojai:
> Bring it as close as possible to the mockups right?

Yeah, but for the popover I don't think we need the selection check, since the button is already expressing the selected lang.

fushinari:
> Yeah, but for the popover I don't think we need the selection check, since the button is already expressing the selected lang.

Oh damn. Had just finished implementing it. :laughing: 

image

fushinari:
> Looks cool, so let's keep it sweat_smilesweat_smile

Wait really? :sweat_smile: I was gonna chuck it off because I found some weird bugs with it.

rafaelmardojai:
> Wait really? 😅 I was gonna chuck it off because I found some weird bugs with it.

That's why I avoided it 😂, so I will leave to you the choice.

fushinari:
Okay, i think the bug might be simple. I'll keep it if I can fix them reasonably quickly. Btw, I changed some of the design into this:
image

I'm a bit conflicted because this looks closer to the mockups but might have been better as it was earlier.

rafaelmardojai:
Also the mockups are just reference. The closeness depends on the real world  issues 😂😅

fushinari:
> Looks good. But in what it difer from the previous implementation? Have you removed the recent langs list?

It's there. Just tiny adjustments. Am a bit busy with other stuff too at the moment so work's going a tad slow. So, I've pushed the minor changes for now.

fushinari:
I just spent nearly 30 minutes debugging something only to find it's a GTK bug: https://gitlab.gnome.org/GNOME/gtk/-/issues/1779

fushinari:
image

We have a bug somewhere...

fushinari:
Current UI during translation:
image

I honestly love this entire design!

rafaelmardojai:
Yeah, looks awesome!

Could we make the whole right side insensitive during translation process?

rafaelmardojai:
You now why the popover slows down when there's a text translated?
Probably related to the recent list redrawing, but why only when a text has been translated?

I hope it's noticeable in this GIF:
Peek 06-10-2020 18-34

Note: We should clear the search entry when the popover is closed.

rafaelmardojai:
I think it's related to the permanent thread. But why?

rafaelmardojai:
Printing anything in the `while` fixes the issue. 😕😕

```python
while True:
    print('why?')
    if self.trans_queue:
        ...
```

rafaelmardojai:
Changing the thread to this should fix the issue:
```python
def run_translation(self):
    while self.trans_queue:
        # Code ...
    self.active_thread = None
```

But, so we will be having the virtual issue again?

fushinari:
> But why?

Probably because an if condition is constantly being checked. But that's in another thread, it shouldn't be impacting the GUI...

> Printing anything in the while fixes the issue.

That's really really odd...

> Changing the thread to this should fix the issue:
>
> But, do we will be having the virtual issue again?

If we do this, won't the thread need to be started again whenever there's another translation? That might be an even bigger hit in performance.

fushinari:
Btw, what do you mean by virtual issue?

fushinari:
I can't decide if this is better or worse...
image

rafaelmardojai:
> Btw, what do you mean by virtual issue?

The issue we wanted to fix with the permanent thread. But forget it, I don't think this is the case.

> I can't decide if this is better or worse...

What do you mean?

fushinari:
> What do you mean?

I added 2px extra top and bottom margins for the labels. I feel like there's wasted space now...

rafaelmardojai:
> I added 2px extra top and bottom margins for the labels. I feel like there's wasted space now...

Yeah, and the fact that we have a long list neither helps.

rafaelmardojai:
The clear action should also clear the right text and save all to history? 🤔

rafaelmardojai:
> The clear action should also clear the right text and save all to history?

Not really, I was writing half sleep. 😅😅

fushinari:
> Yeah, and the fact that we have a long list neither helps.

Yes, it looks neater when you look at it for a while but really, do we need that padding?

Btw, is the defaulting to Auto intended?

rafaelmardojai:
> Yes, it looks neater when you look at it for a while but really, do we need that padding?

Not really, it also looks good without the padding.

> Btw, is the defaulting to Auto intended?

No, it was the default before, so I keep it. And I don't really have a hard opinion about it.

fushinari:
Looking into the history duplication issue... What's actually happening is that whenever `ui_return` is triggered, it goes into `history_update` and then somehow triggers `text-changed` signal twice... We can work around it ofc, but maybe we should instead try to find out what exactly is happening here.

fushinari:
The history part is a rabbit hole full of issues... Might just rewrite that tomorrow if I get time...

rafaelmardojai:
I have pushed my responsive design experiment. The only issue I'm having is that the window doesn't shrink even when it has space. Any ideas? What I'm missing?

rafaelmardojai:
`700` px of width for the responsive rule is probably to much right?

rafaelmardojai:
`700` is fine, otherwise the UI wouldn't change when a language with long name is selected.

rafaelmardojai:
> I have pushed my responsive design experiment. The only issue I'm having is that the window doesn't shrink even when it has space. Any ideas? What I'm missing?

Sorry for being stupid sometimes, I just forgot to wrap the label in the network error view. 😂😂

rafaelmardojai:
The mobile version is looking like this:
Captura de pantalla de 2020-10-07 21-39-54

fushinari:
Looks great so far!

I found even more bugs with the history system... It has several types of crashes so far. Some even involving the X server. I have no idea what to do except rewrite. But i think I'll do that later.

fushinari:
I had reimplemented the history. It's mostly the same but with some tiny but huge (if that makes any sense) changes that should solve all the issues we're having. But here's the problem... I'm stuck at the loading screen for some reason...

fushinari:
Okay, it's not any changes I made... It's just eternally in the loading screen, which brings my attention to another matter, we need a timeout... What's happening is this: The line
```python
self.lang_speech = list(lang.tts_langs(tld='com').keys())
```
is simply not returning anything. Instead it just hangs. Quite possibly, it's just that my IP has been banned or something of the sort.

rafaelmardojai:
The TTS isn't even working, so I don't know if we should remove it at the end.

rafaelmardojai:
And if we keep TTS we should port it from pydub to GStreamer.

fushinari:
> The TTS isn't even working, so probably we should remove it at the end.

Actually, the TTS was working fine for me. But launching the application itself is a hit or miss.

fushinari:
After I finished the original rewrite of the history system, I realized it was still not working right. As I said earlier above:

> Looking into the history duplication issue... What's actually happening is that whenever ui_return is triggered, it goes into history_update and then somehow triggers text-changed signal twice... We can work around it ofc, but maybe we should instead try to find out what exactly is happening here.

Then once I worked that around (using a variable `history_traverse`), I found that many cases weren't handled properly yet. So, I added the needed handling.

But then I found a possibly better implementation. We can avoid the workaround too. The GtkTextBuffer signal `end-user-action` can be used. But I don't really know if it comes with any disadvantages so it might need more testing. Actually, all of it needs more testing. I'm surprised I didn't notice until just now exactly how broken the original history system was.

fushinari:
And with that, I'd say our work is almost done.

rafaelmardojai:
Nice! We should give us some time of testing and then we can merge this!

src/main.py
```diff
@@ -40,8 +41,8 @@ def do_startup(self):
         GLib.set_prgname('com.github.gi_lom.dialect')
         self.setup_actions()
 
-        # Init Handy
-        Handy.init()
+        Handy.init()  # Init Handy
+        Gst.init(None)  # Init Gst
```
rafaelmardojai, line 45: We should init Gst right after the `gi.repository` import or it will fail in some setups.

fushinari: Oh? Are you sure? What's an example of such a setup? Afaik, that shouldn't be happening...

rafaelmardojai: I don't really know what comes to play to make that happen, but I had this issue in Blanket with a Manjaro user: https://github.com/rafaelmardojai/blanket/issues/6.

Lollypop for example, does the same: https://gitlab.gnome.org/World/lollypop/-/blob/master/lollypop/application.py#L20

fushinari: That shouldn't be impacting Dialect since there's no chance at all that we will touch Gst until after do_startup. In Blanket and in Lollypop, you're actually using it before the application startup is complete. (For example, in Blanket, during __init__)

rafaelmardojai: Yeah, you are right!

fushinari:
I think we need to kill the "Listen" feature. It works but I almost wish it didn't. The issue is that instead of speaking in a voice suitable for whatever language is selected, it speaks in the English US voice using the transliterated text. (For eg, in my language, Malayalam.)

src/window.py
```diff
@@ -0,0 +1,547 @@
...
+            self.left_lang_label.set_label(LANGUAGES[code].capitalize())
+            # Updated saved left langs list
+            if code in self.left_langs:
+                # Bring lang to the top
+                index = self.left_langs.index(code)
+                self.left_langs.insert(0, self.left_langs.pop(index))
+            else:
+                self.left_langs.pop()
+                self.left_langs.insert(0, code)
+            self.settings.set_value('left-langs',
```
fushinari, line 251: This is making it really really slow to switch...

rafaelmardojai: Yeah, we could save that when the window is closed.

rafaelmardojai:
Should we also show the selected lang in the recent list?

src/window.py
fushinari: Yeah, I've switched to doing that.

fushinari:
Okay, only one bug left afaik.

rafaelmardojai:
We also save the left langs in the translation thread. We will keep that?

fushinari:
> We also save the left langs in the translation thread. We will keep that?

I have mixed feelings about that one, so I'm experimenting on its reliability.

fushinari:
I said one bug, but there are two major bugs:
1. When switching between desktop mode and mobile mode, the app often gets stuck in a weird in-between state.
Screenshot from 2020-10-09 21-01-54

2. When resized too quickly into a way too small size, the window disappears. It's still there, you just won't be able to see it.
dialect

rafaelmardojai:
Yeah, I was able to reproduce the first yesterday.

The second looks very weird, I will try to reproduce it when I'm back at my computer.

rafaelmardojai:
I don't know if the first can be fixed changing the implementation from the GtkActionBar to a GtkRevealer with a GtkBox using custom CSS.

About the second we need to investigate, it can also be a libhandy bug.

fushinari:
I've been thinking. How about we stop using a loading screen and waiting specifically for the TTS? We could just use the already present Spinner and set a tooltip for voice_btn during the loading. This way, we could cut down on the loading times a lot.

fushinari:
> About the second we need to investigate, it can also be a libhandy bug.

I'm pretty sure it is. It seems similar to the modal drag bug that was present in earlier libhandy1 versions.

fushinari:
> About the second we need to investigate, it can also be a libhandy bug.

Weird thing is, you can't reproduce it in the "Loading" stack page.

fushinari:
Right now, I'm very hesitant to play with the GUI  because I use Glade 3.38 and it breaks compatibility with 3.22, which is what you seem to be using. So, every change I've made so far in the GUI (those minor stuff) have been in the XML directly. Even the Glade flatpak seems to be 3.36. So I have no idea what to do next.

rafaelmardojai:
> Right now, I'm very hesitant to play with the GUI  because I use Glade 3.38 and it breaks compatibility with 3.22, which is what you seem to be using. So, every change I've made so far in the GUI (those minor stuff) has been in the XML directly. Even the Glade flatpak seems to be 3.36. So I have no idea what to do next.

I used GNOME Builder buildin Glade to generate some parts in the beginning. But I prefer to edit the XML files directly, and that's what I have been doing 😅😂

fushinari:
> > Right now, I'm very hesitant to play with the GUI  because I use Glade 3.38 and it breaks compatibility with 3.22, which is what you seem to be using. So, every change I've made so far in the GUI (those minor stuff) has been in the XML directly. Even the Glade flatpak seems to be 3.36. So I have no idea what to do next.
> 
> I used GNOME Builder buildin Glade to generate some parts in the beginning. But I prefer to edit the XML files directly, and that's what I have been doing sweat_smilejoy

So... me editing it with Glade 3.38 is fine?

rafaelmardojai:
> So... me editing it with Glade 3.38 is fine?

Yeah, I also don't have the libhandy catalog installed so... 😅😅

The only issue I have with Glade is the unnecessary packaging properties it add to the GtkBox children 😅

fushinari:
> The only issue I have with Glade is the unnecessary packaging properties it add to the GtkBox children sweat_smile

I agree but it makes things somewhat easier if you ignore the issues. I also hate how it doesn't support the "margin" property.

rafaelmardojai:
> I agree but it makes things somewhat easier if you ignore the issues. I also hate how it doesn't support the "margin" property.

Yeah, and we can always cleanup the XML later.

fushinari:
I'll be sending in a huge commit soon.

fushinari:
That fixes one major bug (above mentioned no1) and introduces a minor warning:
```
(com.github.gi_lom.dialect:43889): Gtk-WARNING **: 23:32:27.010: Negative content width -19 (allocation 1, extents 10x10) while allocating gadget (node box, owner GtkBox)
```
[This](https://gitlab.gnome.org/GNOME/gtk/-/issues/1057) issue seems relevant. 

The biggest change is, we're not waiting for gTTS loading anymore. This hugely improves startup time and it also solves many bug risks. As a side effect, the lovely stack pages had to be removed since they're not in use anymore.

You were right, switching away from GtkActionBar to a GtkRevealer fixed the first of the weird bugs.

I also added a basic in-app notification system to solve some of the error-presenting issues we're having. It might need some more work, especially to maybe try and workaround the above-mentioned GTK bug. (We've been doing that a lot haven't we? Here's to hoping GTK4 makes it all better.)

I also confirmed that the second bug we discussed is Wayland-only.

The code right now isn't as clean as I'd like it to be but I hope to fix that later rather than now. I'm too tired for now. :sweat_smile: Gonna take a good long sleep now

fushinari:
Bigger question is, are all the changes for the better or am I missing something? Asking because I probably am :sweat_smile:

rafaelmardojai:
> That fixes one major bug (above mentioned no1) and introduces a minor warning:

Yeah, looks annoying, but at least is not a critical 😂

> The biggest change is, we're not waiting for gTTS loading anymore. This hugely improves startup time and it also solves many bug risks. As a side effect, the lovely stack pages had to be removed since they're not in use anymore.

Looks awesome! The only issue is that the Listen button becomes sensitive when the loading is finished.

> I also confirmed that the second bug we discussed is Wayland-only.

I was suspecting that. I use X.Org and wasn't able to reproduce the bug. 

There's also this bug:

Captura de pantalla de 2020-10-10 21-02-57

It's related to the selector popover begin set to the mobile mode buttons. This was fixed by the `mobile_mode = None` thing that was helping to setup everything correctly depending on the window size 😂. This can be fixed by calling `self.toggle_mobile_mode` at the end of `self.setup`.

Also, what's the difference from using `threading.Timer` or `GLib.timeout_add_seconds`?

rafaelmardojai:
Why are we doing the `second_buffer = self.right_buffer` thing instead of calling `self.right_buffer` directly?

rafaelmardojai:
Fixed the Listen button and popover issues.

rafaelmardojai:
More issues:

- If you close the window when the `load_lang_speech` thread is still running, the app will not terminate until the tread is done. Not a real "issue" but there is it. 😂
- [ ] The Copy and Listen buttons will become sensitive with the right text empty, because we make those sensitive in the `translation` function without checking.
- [ ] When the network connection is recovered, you can't translate the failed translation since it was already "translated" 😂.

fushinari:
> This can be fixed by calling self.toggle_mobile_mode at the end of self.setup.

I was planning to add it in anyway because I don't know if check-resize will be called when a mobile device forces its geometry on the app. (I don't know because I don't own the devices to test :sweat_smile: )

> Also, what's the difference from using threading.Timer or GLib.timeout_add_seconds?

I think they could work similarly if we make the function return False, but I simply prefer this. Most importantly, idle_add from the main thread has always resulted in slight performance drops so far. So, I'm trying to avoid that.

> Fixed the Listen button and popover issues.

Nice!

> Why are we doing the `second_buffer = self.right_buffer` thing instead of calling `self.right_buffer` directly?

Just remnants of the old code. Safely removable.

> If you close the window when the load_lang_speech thread is still running, the app will not terminate until the tread is done. Not a real "issue" but there is it.

Threads, so yeah. We'll let python do the garbage collection and such there. Forcing it to terminate it is not an option anyway.

>  The Copy and Listen buttons will become sensitive with the right text empty, because we make those sensitive in the translation function without checking.

Yes, I noticed immediately after pushing yesterday. :laughing: 

> When the network connection is recovered, you can't translate the failed translation since it was already "translated".

On a similar note, if the user gets back connectivity after `load_lang_speech` fails, he has no way to try again.

fushinari:
Plenty of rough edges, I know. Just wanted to make the necessary changes asap and then work on the issues that come with.

rafaelmardojai:
> Plenty of rough edges, I know. Just wanted to make the necessary changes asap and then work on the issues that come with.

Yes, also most of the issues were there before 😂

fushinari:
> Yes, also most of the issues were there before joy

Or we had bigger issues in their place :rofl: I mean, we didn't have any feedback at all even when translation fails entirely

fushinari:
> On a similar note, if the user gets back connectivity after load_lang_speech fails, he has no way to try again.

rafaelmardojai:
> How do we even solve this?

That's a tricky one.

I can think in having something that checks network connection in a time cycle when an error has been raised. But I don't have it clear.

fushinari:
> I can think in having something that checks network connection in a time cycle when an error has been raised. But I don't have it clear.

Keep a permanent thread (the translation thread isn't permanent anymore btw) that checks for basic connectivity and then, if basic connectivity is found tries to load the languages list again?

rafaelmardojai:
> Keep a permanent thread (the translation thread isn't permanent anymore btw) that checks for basic connectivity and then, if basic connectivity is found tries to load the languages list again?

Yeah, something like that.

rafaelmardojai:
I have been trying multiple ways to detect if there's connectivity, but all those keep failing when you start the app without connectivity. Also makes the translation fail even if the connection has comeback.

Current attemp:
```python
import socket
import time

while True:
    try:
        host = socket.gethostbyname('www.google.com')
        sock = socket.create_connection((host, 80), 2)
        sock.close()
        break # if we only want to call it once. 

    except Exception as exc:
        # print
    finally:
        time.sleep(5)
```

Ideas?

fushinari:
> I have been trying multiple ways to detect if there's connectivity, but all those keep failing when you start the app without connectivity. Also makes the translation fail even if the connection has comeback.

Sorry, but I probably won't be able to work on this for a few days, again. 

I've seen this being done before:
```python
try:
    socket.setdefaulttimeout(10)
    socket.socket(socket.AF_INET, socket.SOCK_STREAM).connect(('8.8.8.8', '53'))
    return True
except socket.error as exc:
    print(exc)
    return False
``` 

That should be the fastest method.

Also, NetworkManager provides connectivity info on its DBUS API through `NMConnectivityState`. Don't know if that's reliable enough though...

> Also makes the translation fail even if the connection has comeback.

Wait how? That shouldn't be happening...

fushinari:
How about this? Even if the last attempt of whatever we're trying to do (translation or audio) failed, we keep the button sensitive, but when running the next time after a failure, we check for connectivity. So, if the last time failed, we'll have the warning icon feedback but not making it insensitive. In the case of audio, if lang list loading failed, we try to load it again, wait for that and then try to produce audio. This way, almost all cases should be handled.

rafaelmardojai:
> How about this? Even if the last attempt of whatever we're trying to do (translation or audio) failed, we keep the button sensitive, but when running the next time after a failure, we check for connectivity. So, if the last time failed, we'll have the warning icon feedback but not making it insensitive. In the case of audio, if lang list loading failed, we try to load it again, wait for that and then try to produce audio. This way, almost all cases should be handled.

I'm with you, that looks like the best approach for us!

fushinari:
Screenshot from 2020-10-12 11-25-47

For all the hate Ubuntu gets, it's probably still the most polished-looking linux distro out there. Everything just looks right... Green is obviously and logically "suggested-action" for example. Sure it might break with more CSS, but for now, it's damn good. Also, the Yaru icon theme's monochrome icons are so damn pretty and they make more logical sense than Adwaita's icons. For example, the Sound OSD icon when going over the limit, the microphone icon, etc. Overall, I can see where their design team's efforts are going.

On a sidenote, is their any way to make GNOME behave like Plasma where scrolling on the microphone icon will increase microphone volume? Right now, scrolling on the microphone icon increases speaker volume... GNOME lacks polish in some of the things I traditionally expect it to be polished. (But it's polished in literally every other aspect...)

Don't know why I'm ranting about that here... But anyway, I tested the Ubuntu-like setup (with Yaru 20.10) to check if what you said might be true, about the horizontal flip icon being different. I wanted to ensure it wasn't different in at least the most used Linux distro. I also wanted to check how badly Yaru broke CSS. But turns out they don't break much ever since their switch to Adwaita base if you're using the expected versions of each component (like libhandy).

fushinari:
> I'm with you, that looks like the best approach for us!

I just realized... If we're doing this, we don't really need a connectivity check at all, do we?

rafaelmardojai:
> I just realized... If we're doing this, we don't really need a connectivity check at all, do we?

Nope, we don't need it.

fushinari:
> Nope, we don't need it.

Could you check if everything behaves as expected now?

rafaelmardojai:
When the Listen button fails it becomes insensitive. 
Also the connection will never recover if you start the app without connection, I feel it's some `urllib3` issue.

rafaelmardojai:
> For all the hate Ubuntu gets, it's probably still the most polished-looking linux distro out there. Everything just looks right... Green is obviously and logically "suggested-action" for example. Sure it might break with more CSS, but for now, it's damn good. Also, the Yaru icon theme's monochrome icons are so damn pretty and they make more logical sense than Adwaita's icons. For example, the Sound OSD icon when going over the limit, the microphone icon, etc. Overall, I can see where their design team's efforts are going.

Yeah, the app looks really nice there!

> On a sidenote, is their any way to make GNOME behave like Plasma where scrolling on the microphone icon will increase microphone volume? Right now, scrolling on the microphone icon increases speaker volume... GNOME lacks polish in some of the things I traditionally expect it to be polished. (But it's polished in literally every other aspect...) (edit: i opened an issue for this)

I never tried that feature in the microphone icon, but I supposed that it worked as the volume one. At least I would expect it to do nothing, instead of increasing speaker volume. 😂

fushinari:
> When the Listen button fails it becomes insensitive.

Listen button fails? Sorry, i don't understand. When you press the listen button, `ui_voice` is called. This will make the spinner insensitive for the duration of the action. Once the words have been spoken or it fails, it should go back to normal if other conditions are met. (second_text not being empty and the selected language being supported)

> Also the connection will never recover if you start the app without connection, I feel it's some `urllib3` issue.

Wait really? How exactly do I reproduce this? It seems to work fine for me.

> Edit: Super minor issue: `run_translation` sends a notification with a space after the `\n`.

Solved.

rafaelmardojai:
> Listen button fails? Sorry, i don't understand. When you press the listen button, ui_voice is called. This will make the spinner insensitive for the duration of the action. Once the words have been spoken or it fails, it should go back to normal if other conditions are met. (second_text not being empty and the selected language being supported)

I mean, when the voice download fails, the Listen button becomes insensitive, so you can't retry.
Peek 12-10-2020 09-48

rafaelmardojai:
> Wait really? How exactly do I reproduce this? It seems to work fine for me.

1. Disable internet connection from computer
2. Start Dialect
3. `load_lang_speech` will fail and console will print: 

    ```bash
    Error: Unable to get language list: HTTPSConnectionPool(host='translate.google.com', port=443): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7fd8742f17d0>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution'))
    ```
    Witch is an strange error. Probably related to something else in my side.

4. Enable internet connection again
5. Dialect can't even translate and `run_translation` keeps failing.

rafaelmardojai:
Also if you translate something and `load_lang_speech` is still loading, the Listen button will become sensitive. We probably need some `voice_loading` var to check.

Clicking the Listen button  in that state will not fail, but will start `load_lang_speech` again.

fushinari:
>     1. Disable internet connection from computer
> 
>     2. Start Dialect
> 
>     3. `load_lang_speech` will fail and console will print:
>        ```shell
>        Error: Unable to get language list: HTTPSConnectionPool(host='translate.google.com', port=443): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7fd8742f17d0>: Failed to establish a new connection: [Errno -3] Temporary failure in name resolution'))
>        ```
>        
>        
>        Witch is an strange error. Probably related to something else in my side.
> 
>     4. Enable internet connection again
> 
>     5. Dialect can't even translate and `run_translation` keeps failing.

Oh... this sounds like a DNS issue and it could be that I can't reproduce it because I use dnscrypt-proxy all the time...

fushinari:
> Also if you translate something and `load_lang_speech` is still loading, the Listen button will become sensitive. We probably need some `voice_loading` var to check.
> 
> Clicking the Listen button in that state will not fail, but will start `load_lang_speech` again.

Yeah, I messed this up badly. I found out only now...

fushinari:
@rafaelmardojai Could you check now? I should've covered most cases by now.

fushinari:
>     5. Dialect can't even translate and `run_translation` keeps failing.

I kept on trying to reproduce this. No dice. It just works as intended on my system:
```
Error: Unable to get language list: HTTPSConnectionPool(host='translate.google.com', port=443): Max retries exceeded with url: / (Caused by NewConnectionError('<urllib3.connection.HTTPSConnection object at 0x7fd672200880>: Failed to establish a new connection: [Errno 101] Network is unreachable'))
```

rafaelmardojai:
> I kept on trying to reproduce this. No dice. It just works as intended on my system:

That's the error expected, I don't know what I'm getting the `Temporary failure in name resolution` DNS thing.

rafaelmardojai:
> Could you check now? I should've covered most cases by now.

All looks good!

fushinari:
> That's the error expected, I don't know what I'm getting the `Temporary failure in name resolution` DNS thing.

Yeah, nothing we can do about a bug in googletrans or urllib anyway... You could test how exactly it happens and report it to googletrans or urllib.

> All looks good!

Good enough to review and merge? (After some cleanup ofc)

fushinari:
Technically, we'd already accomplished the original goal of the PR a long time ago lol

rafaelmardojai:
> Technically, we'd already accomplished the original goal of the PR a long time ago lol

😅😅😂

Yes, some cleanup and we are ready!

rafaelmardojai:
With this merged, we only need to find a new icon and we can release!

fushinari:
> With this merged, we only need to find a new icon and we can release!

Yeah.

There's something I'm conflicted about. `self.run_translation` is called in a daemon thread. It's the reason why Dialect doesn't exit immediately most of the time. But I'm not sure I want this behavior to change. For example, what if it exits while the history entry is being added?

fushinari:
I renamed "first" and "second" to "src" and "dest" for clarity... But now I'm thinking maybe I should've only changed it for src_language and dest_language, and not src_text and dest_text.

Also, should we be renaming left_buffer and right_buffer? I mean... they aren't really just left and right anymore are they?

rafaelmardojai:
> Also, should we be renaming left_buffer and right_buffer? I mean... they aren't really just left and right anymore are they?

Yes, if we're changing the left/right thing, we should change it every where.

fushinari:
Exiting should now be instant under normal circumstances.

fushinari:
What actually changed is that with daemon=True gone, the thread now exits by itself only when its job is complete. I don't know if that's better or worse tbh...

fushinari (approved):

fushinari:
And with that last bit of missing functionality added in, I think we're done! :tada: Please give a final review and merge. :)

rafaelmardojai:
> What actually changed is that with daemon=True gone, the thread now exits by itself only when its job is complete. I don't know if that's better or worse tbh...

As long as the window closes instantly, we shouldn't care 😂

fushinari:
> As long as the window closes instantly, we shouldn't care joy

True ig :rofl:

rafaelmardojai:
By default we set the same recent lang list for both sides, so the app starts with `English -> English` after installation, that's what we want?

rafaelmardojai:
Also when the same lang is select in both sides we do nothing, is fine keep it like that?

rafaelmardojai:
Pushed fixes for the issues commented above. Checkout and merge if is ready!

rafaelmardojai:
When this is merged, I will open a new PR enabling localization. When we release people might want to contribute locals.

rafaelmardojai:
Bug: The char count wouldn't be updated when you use the Clear button.

rafaelmardojai:
We have two minor issues with the live translation that you can see in the GIF.

1. I don't think we can do much to solve this: when you clear the src text by hand, always the last letter is translated. 
2.  When the dest text gets empty, Copy and Listen buttons are sensitive. 

Peek 12-10-2020 17-09

rafaelmardojai:
> 2. When the dest text gets empty, Copy and Listen buttons are sensitive.

Pushed a fix for this.

rafaelmardojai:
When the lang is changed in any selector, translation should be executed? At least in live translation?

fushinari:
> By default we set the same recent lang list for both sides, so the app starts with `English -> English` after installation, that's what we want?

I just found why gi-lom decided on the "Auto" language selection at start. It's from the Google Translate behavior, which is probably the behavior people will expect too. So maybe we should revert back to using Auto?

> When this is merged, I will open a new PR enabling localization. When we release people might want to contribute locals.

Awesome!

> When the lang is changed in any selector, translation should be executed? At least in live translation?

Yeah, let's do that.

> I don't think we can do much to solve this: when you clear the src text by hand, always the last letter is translated.

This is the weirdest bug so far. I thought end-user-action might not be getting emitted properly. But turns out that works fine. Then I looked for other failures. Nope. Everything's fine and `self.dest_buffer.set_text('')` does get executed... I have no idea what's going on here...

fushinari:
I don't understand why it was happening, but now I can pretend I actually fixed it :laughing:

rafaelmardojai:
> I just found why gi-lom decided on the "Auto" language selection at start. It's from the Google Translate behavior, which is probably the behavior people will expect too. So maybe we should revert back to using Auto?

Yeah, I'm fine with that, it's a good default.

PR #45, opened 2020-10-14
Localization support 

rafaelmardojai:
What does this PR?

- Add localization support with gettext
- Improve appstream metadata and .desktop
- Generate `.pot` file for translators ease
- Add Spanish translation
- Also minor bug fix I commited here at the end 😂 

Tested with exported flatpak.


fushinari (approved):
LGTM!

PR #53, opened 2020-10-16
Add Russian translation

nexeleriot:


nexeleriot:
I need to do one more thing

nexeleriot:
There is another PR now

PR #67, opened 2020-10-23
app icon: update for the A -> D change

jimmac:
Fixes https://github.com/gi-lom/dialect/issues/66


fushinari:
Thank you for your work!

Also, it looks like CI is broken @rafaelmardojai

rafaelmardojai:
@fushinari Flatpak is failing because the search provider is installing directly to `/usr/share`.

PR #71, opened 2020-10-25
Support "Did you mean" mistakes suggestion

### LINKED ISSUE #55: "Did you mean" suggestion when misspelled text
UX would be much better and more comparable with web Google Translate interface, when you app would be able to provide "Did you mean" suggestions when text has wrong, misspelled words. I don't know whether API provides this information but IMHO without this kind of suggestion for end users the app has little sense for me.

Zrzut ekranu z 2020-10-17 15-21-14

rafaelmardojai:
Fix #55 

Just discovered that **py-googletrans** returns this data in `extra_data['possible-mistakes']`.

### Demo
Peek 25-10-2020 12-46

### DIFF
--- a/src/window.py
+++ b/src/window.py
@@ -703,7 +703,7 @@ class DialectWindow(Handy.ApplicationWindow):
                     src_language = str(self.translator.detect(src_text).lang)
                     GLib.idle_add(self.src_lang_selector.set_property,
                                   'selected', src_language)
-                    if not src_language in src_langs:
+                    if not src_language in self.src_langs:
                         self.src_langs[0] = src_language
                 except Exception:
                     self.trans_failed = True
--- a/src/window.py
+++ b/src/window.py
@@ -37,6 +37,9 @@ class DialectWindow(Handy.ApplicationWindow):
 
     menu_btn = Gtk.Template.Child()
 
+    mistakes = Gtk.Template.Child()
+    mistakes_btn = Gtk.Template.Child()
+    mistakes_label = Gtk.Template.Child()
     char_counter = Gtk.Template.Child()
     src_text = Gtk.Template.Child()
     clear_btn = Gtk.Template.Child()
@@ -76,6 +79,8 @@ class DialectWindow(Handy.ApplicationWindow):
     # Connectivity issues monitoring
     trans_failed = False
     voice_loading = False
+    # Trans mistakes
+    trans_mistakes = None
 
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
@@ -223,6 +228,8 @@ class DialectWindow(Handy.ApplicationWindow):
         self.paste_btn.connect('clicked', self.ui_paste)
         # Translate button
         self.translate_btn.connect('clicked', self.translation)
+        # Mistakes button
+        self.mistakes_btn.connect('clicked', self.on_mistakes_clicked)
 
         # Right buffer
         self.dest_buffer = self.dest_text.get_buffer()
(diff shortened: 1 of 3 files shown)


src/window.py
```diff
@@ -565,6 +572,12 @@ def update_trans_button(self, button, keyboard):
 
         return Gdk.EVENT_PROPAGATE
 
+    def on_mistakes_clicked(self, _button):
+        self.mistakes.set_reveal_child(False)
+        self.src_buffer.set_text(self.trans_mistakes[1])
+        # Run trasnlation again
+        self.translation(None)
```
rafaelmardojai, line 579: It's fine to run the translation again? Since we are already showing the translation without the mistake.

fushinari: Yes, running the translation again is still a good idea.

fushinari:
Few questions and/or suggestions:
- Wouldn't it have been easier to simply use a GtkInfoBar? This seems like a job specifically suited to it too.
- Shouldn't only the possible mistakes be the buttons? i.e. the "Did you mean: " should just be a label, not a part of the button.
- The "?" at the end should probably be avoided. That's the Google Translate default behavior and avoids unnecessary complications from the src text also ending in "?".

Otherwise, looks amazing! I didn't even know there was an `extra_data` variable. Looks interesting. It gives a lot more goodies too, like "See also", "Synonyms", "Definitions", etc. Nice!

rafaelmardojai:
> Wouldn't it have been easier to simply use a GtkInfoBar? This seems like a job specifically suited to it too.

Yeah, but it always force the buttons to the end.

Captura de pantalla de 2020-10-25 15-06-09

Is that fine?

rafaelmardojai:
This is how looks your suggestions without using GtkInfoBar:
Captura de pantalla de 2020-10-25 15-16-28

rafaelmardojai:
> Yeah, but it always force the buttons to the end.

Now looking it more, probably that way is more balanced?

Captura de pantalla de 2020-10-25 15-32-10

fushinari:
Wouldn't it be better to follow Google Translate behavior and just make it a link? Would look a lot more consistent imo. I feel like buttons aren't suited for the job.

rafaelmardojai:
> Wouldn't it be better to follow Google Translate behavior and just make it a link? Would look a lot more consistent imo. I feel like buttons aren't suited for the job.

Yeah, I wanted that too. But I was lazy to implement the custom CSS, since `GtkLinkButton` is supposed to open uris.

rafaelmardojai:
Also that was the reason to use a flat button in the beginning, but looks like it not work for buttons inside a `GtkButtonBox` (that is what uses `GtkInfoBar`).

rafaelmardojai:
@fushinari Same here. Here we only need to finish the appearance. 
If you have time, feel free to push.

fushinari:
> Yeah, I wanted that too. But I was lazy to implement the custom CSS, since `GtkLinkButton` is supposed to open uris.

Luckily, we don't need to. :)

Screenshot_20201026_111744

Looks like it works. Might need some more testing. py-googletrans or the site itself is doing the escaping for us, so we don't even need that.

fushinari:
Oh yeah, also the above image is an example for why we definitely need to keep the re-translation. When only "dit" is given, the translation done is actually for the word "ditch" even though the suggestion is "did".

fushinari:
Screenshot_20201026_114915

Should I add that icon there? And if I do, which icon is more suitable `error-correct-symbolic` as seen in image or `tools-check-spelling-symbolic` which looks like this: 
image

rafaelmardojai:
I thinks `error-correct-symbolic` is the correct semantically. But `tools-check-spelling-symbolic` looks cooler 😂

fushinari (approved):
Exactly my feelings. :rofl:  Well, let's leave it at the semantically correct thing then. 

Have you tested it? For me, it seems to work fine now. If all else looks okay, go ahead and merge.

TomaszGasior:
Great work! :)

PR #77, opened 2020-10-28
Updates for Turkish language.

v30xy:


fushinari:
Oof, I forgot to wait for CI. But well, I guess it's fine.

rafaelmardojai:
@v30xy You also need to update the other strings marked with `fuzzy` that where auto filled in the `po` files updating.

rafaelmardojai:
> You also need to update the other strings marked with fuzzy that where auto filled during the po files updating.

If you are using a GUI, those strings should be marked with a different color.

fushinari:
@v30xy Sorry, that was my bad, but could you send another PR with those changes?

PR #87, opened 2020-10-29
Ctrl + Q to quit the app

nahuelwexd:
This PR adds a handy shortcut to quit the app, following the recommendations at https://developer.gnome.org/hig/stable/keyboard-input.html

Dunno if this should be accompanied by the addition of the keyboard shortcuts window

### DIFF
--- a/src/main.py
+++ b/src/main.py
@@ -89,6 +89,11 @@ class Dialect(Gtk.Application):
         about_action.connect('activate', self.on_about)
         self.add_action(about_action)
 
+        quit_action = Gio.SimpleAction.new('quit', None)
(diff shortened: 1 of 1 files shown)


fushinari:
Locally, I already have the Keyboard Shortcuts window ready. The reason I'm not pushing it into master right now is because I don't want to add more translation work right before 1.1.0.

Thanks for your work!

nahuelwexd:
> Locally, I already have the Keyboard Shortcuts window ready. The reason I'm not pushing it into master right now is because I don't want to add more translation work right before 1.1.0.

Okay, makes sense, great then :D

PR #92, opened 2020-11-01
Add a shortcuts window

fushinari:

PR #94, opened 2020-11-02
Update Portuguese translation

sirguxx:
Apparently, the translation was not updated automatically like the others after PR #92


fushinari:
Thanks for the work!

PR #99, opened 2020-11-03
Update Persian translation

musicsd:


po/fa.po
```diff
@@ -157,6 +178,7 @@ msgstr "ترجمه"
 #: data/resources/shortcuts-window.ui:31
 msgctxt "shortcuts window"
 msgid "Quit"
+msgstr "برونرفت"
 msgstr ""
```
rafaelmardojai, line 182: Remove this empty `msgstr`.

PR #105, opened 2020-12-08
Update it.po

albanobattistella:


po/it.po
```diff
@@ -203,12 +197,12 @@ msgstr "Preferenze"
 #, fuzzy
```
fushinari, line 197: Could you remove the fuzzy marker here too since you've manually translated it?

albanobattistella: Hi,

done!

fushinari:
Thank you for the work!

PR #109, opened 2020-12-12
Pronunciations

### LINKED ISSUE #108: Show pinyin for Chinese translations
For Chinese language learners the Pinyin that shows up underneath the characters can be very helpful. Does the api expose that at all? It is shown in the below image `nǐ hǎo`. Thanks for creating this fantastic application!

image

fushinari:
Fixes #108 

This adds pronunciations for every language. Not just Chinese.

### DIFF
--- a/data/resources/style.css
+++ b/data/resources/style.css
@@ -12,6 +12,13 @@
   border-right: 1px solid @borders;
 }
 
+.pronunciation {
+  border-top: 0;
+  border-left: 1px solid @borders;
+  border-right: 1px solid @borders;
+  border-bottom: 1px solid @borders;
+}
+
 .translation-frame  {
   background: @theme_base_color;
   border-top-left-radius: 9px;
--- a/data/resources/style.css
+++ b/data/resources/style.css
@@ -13,10 +13,8 @@
 }
 
 .pronunciation {
+  border: 1px solid @borders;
   border-top: 0;
-  border-left: 1px solid @borders;
-  border-right: 1px solid @borders;
-  border-bottom: 1px solid @borders;
 }
 
 .translation-frame  {
--- a/data/resources/window.ui
+++ b/data/resources/window.ui
@@ -260,7 +260,7 @@
                               <object class="GtkRevealer" id="pronunciation_revealer">
                                 <property name="visible">True</property>
                                 <property name="can-focus">False</property>
-                                <property name="transition-type">none</property>
+                                <property name="transition-type">slide-down</property>
                                 <property name="reveal-child">False</property>
                                 <child>
                                   <object class="GtkScrolledWindow">
@@ -290,9 +290,6 @@
                                   </object>
                                 </child>
                               </object>
-                              <packing>
-                                <property name="expand">False</property>
-                              </packing>
                             </child>
                             <child>
                               <object class="GtkInfoBar" id="mistakes">
--- a/data/resources/window.ui
+++ b/data/resources/window.ui
@@ -278,7 +278,6 @@
                                         <property name="xalign">0</property>
                                         <property name="wrap">True</property>
                                         <property name="wrap-mode">word-char</property>
-                                        <property name="label" translatable="yes">~</property>
                                         <style>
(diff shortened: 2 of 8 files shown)


fushinari:
This was easy to implement but needs some polish.

One-line text is fine:
Screenshot_20201212_110236

Multiple lines is not. We need scrolling because too much text will force the app UI to scale. But using a GtkScrolledWindow forces the scrollbar to appear even if there are only two lines of text:
Screenshot_20201212_111622

@rafaelmardojai what do you think?

rafaelmardojai:
I think this feature should be hidden with a preference, probably available in the menu. 

> Multiple lines is not. We need scrolling because too much text will force the app UI to scale. But using a GtkScrolledWindow forces the scrollbar to appear even if there are only two lines of text:

Yeah, that doesn't look good. Probably could help if it expand to a certain height, but I don't know if is possible.

data/resources/style.css
```diff
@@ -12,6 +12,13 @@
   border-right: 1px solid @borders;
 }
 
+.pronunciation {
+  border-top: 0;
+  border-left: 1px solid @borders;
+  border-right: 1px solid @borders;
+  border-bottom: 1px solid @borders;
```
rafaelmardojai, line 19:
```suggestion
  border: 1px solid @borders;
  border-top: 0;
```

fushinari:
> I think this feature should be hidden with a preference, probably available in the menu.

You mean, as a checkbox in the popover menu?

> Yeah, that doesn't look good. Probably could help if it expand to a certain height, but I don't know if is possible.

I don't think that's possible but I'm not sure. Lemme see

rafaelmardojai:
> You mean, as a checkbox in the popover menu?

Yes

fushinari:
Sorry, I couldn't find time today. I'll do it later tomorrow evening (relatively speaking, around 15 hrs from posting this). But if you find time to push in the meantime, feel free to do so

rafaelmardojai:
I implemented the setting, but couldn't find a way to solve the text display issues 😕.

data/resources/window.ui
```diff
@@ -256,6 +256,41 @@
...
+                                        <property name="visible">True</property>
+                                        <property name="can-focus">False</property>
+                                        <property name="margin-top">4</property>
+                                        <property name="margin-bottom">4</property>
+                                        <property name="margin-start">9</property>
+                                        <property name="margin-end">9</property>
+                                        <property name="xalign">0</property>
+                                        <property name="wrap">True</property>
+                                        <property name="wrap-mode">word-char</property>
+                                        <property name="label" translatable="yes">~</property>
```
rafaelmardojai, line 281: Any reason for keeping the label property?

fushinari: Oops. Just a remnant from testing. Removing it is fine.

fushinari:
> I implemented the setting, but couldn't find a way to solve the text display issues 😕.

I see, that's too bad but since it's an optional feature and there's nothing we can really do about it, I guess it's fine? @cwmke is that an okay trade-off?

rafaelmardojai:
Note: I have updated all the python deps, and looks like TTS is working again. No code update was needed on our side, so I will update the deps also for Flathub.

cwmke:
> > I implemented the setting, but couldn't find a way to solve the text display issues 😕.
> 
> I see, that's too bad but since it's an optional feature and there's nothing we can really do about it, I guess it's fine? @cwmke is that an okay trade-off?

Seems fine to me. Just having the text there will already eliminate my need for Google Translate’s web app on my end.

Great job you two! This is the fastest turnaround on an issue I have ever experienced. 😃

fushinari:
> so I will update the deps also for Flathub.

Works fine on my end, go ahead. I've merged this since we seem to have accomplished what we wanted to and it looks good.

fushinari:
> Seems fine to me. Just having the text there will already eliminate my need for Google Translate’s web app on my end.
> 
> Great job you two! This is the fastest turnaround on an issue I have ever experienced.

Really glad to know!

PR #111, opened 2020-12-17
Update Persian translation

musicsd:


fushinari:
As always, thank you for always keeping it up-to-date!

PR #118, opened 2021-01-19
L10n: Fix minor spelling errors in Turkish

v30xy:


fushinari:
As always, thanks for the work!

Issue #112, opened 2020-12-23
Argos Translate Backend


jannuary:
[Argos Translate](https://github.com/argosopentech/argos-translate) is an offline translation app/library written in Python. 

It would be great to have that as an option instead of sending everything through Google.

rugk:
Aka https://translate.astian.org/. There you would have a nice API endpoint for that.

fossdd:
[Argos Translate](https://github.com/argosopentech/argos-translate) is a "library" used by [LibreTranslate](https://github.com/uav4geo/LibreTranslate).

[`https://translate.astian.org/`](https://translate.astian.org/) and [`https://libretranslate.com/`](https://libretranslate.com/) are instances of LibreTranslate.

LibreTranslate has a great, simple API. [See docs](https://libretranslate.com/docs)

Issue #114, opened 2021-01-12
[Feature request] Allow other online sources than Google translate


Nokia808:
Hi. 

Currently, your cute application allow translation from only one online site, Google translate. But there are other cute source for online translation & it will be very useful to add an option allowing user to select from several choices.

Example for other online translation sources are:
https://www.bing.com/translator/
https://translate.yandex.com/
https://www.translate.com/

Kindly, see which of them could be integrated with your application.

Best.

fushinari:
There are also other open issues regarding implementing alternative backends. And it's planned. It's just that I haven't had the time to implement it, especially because it needs to be thought out well enough for easy future extension. I assume it's a similar story for @rafaelmardojai .

rugk:
I'd also suggest to open one issue per suggested backend.
There already is #47, #112 and https://github.com/gi-lom/dialect/issues/119.

rafaelmardojai:
We already support Yandex and Bing, does https://www.translate.com/ have any benefit over them?

Issue #47, opened 2020-10-15
Add DeepL backend


ai:
[DeepL](https://www.deepl.com/translator) is insane good in translation quality (much better than Google Translate for many language pairs). Seems like they have [API](https://www.deepl.com/en/docs-api/).

It will be nice to have a setting to switch between Google Translate and DeepL backends.

rafaelmardojai:
To my knowledge you need to pay to have access to the API. You also need a secret KEY to authenticate.

So, If I'm not wrong, we will need a subscription to develop this request. And users aiming to use DeepL, will need to provide their own API KEY.

ai:
Sad.

Good thing, that the basic API key is [free](https://www.deepl.com/en/pro#single).

fushinari:
> Good thing, that the basic API key is [free](https://www.deepl.com/en/pro#single).

No, API access is fully paid sadly.

sp1ritCS:
Too bad that they dont use static pages :)

Request:
```sh
curl 'https://www2.deepl.com/jsonrpc' -H 'Content-type: application/json' --data-raw '{"jsonrpc":"2.0","method": "LMT_handle_jobs","params":{"jobs":[{"kind":"default","raw_en_sentence":"Hello World","raw_en_context_before":[],"raw_en_context_after":[],"preferred_num_beams":4,"quality":"fast"}],"lang":{"source_lang_user_selected":"auto","target_lang":"DE"},"priority":-1,"commonJobParams":{},"timestamp":1602790014301}}'
```
Response:
```
{"jsonrpc":"2.0","result":{"translations":[{"beams":[{"postprocessed_sentence":"Hallo Welt","num_symbols":5,"score":-4997.138279201713,"totalLogProb":5.708958496400825},{"postprocessed_sentence":"Hallo Welt!","num_symbols":6,"score":-4998.529303924396,"totalLogProb":2.932460362221432},{"postprocessed_sentence":"Hallo Welt.","num_symbols":6,"score":-4998.548608063369,"totalLogProb":2.893969321615728},{"postprocessed_sentence":"Hallo, Welt","num_symbols":6,"score":-4998.693697186523,"totalLogProb":2.6046722263858275}],"quality":"normal"}],"target_lang":"DE","source_lang":"IT","source_lang_is_confident":false,"timestamp":1602790398,"date":"20201015"}}
```

If It's ethical / legal to this is up to you / your government.

fushinari:
[...]

PR #120, opened 2021-01-31
Support for multiple backends

fushinari:
Just starting this here so that I remind myself to get back around to it when I can. I think the plan should be to start with #112 (using the [LibreTranslate.com](https://libretranslate.com) instance initially) and then decide where we go from there.

Some decisions need to be made.
- [x] ~Do we simply clear history each time backend is switched or do we maintain separate histories for each backend?~ We clear.
- [x] ~On a similar note, how do we handle saving language preferences? Should we just use some defaults that are available in all services and then reset every time backend is switched or should we maintain separate defaults and separate saved languages?~ Saved per-backend.
- [x] ~What do we do for the UI for switching? Just an HdyComboRow? How badly or how well does it handle long lists? (if we end up implementing too many backends)~ Needs more investigation but let's stick with HdyComboRow for now.
- [x] ~How many backends should be available? How many is too many?~ Again, just start with what we can. So, probably just #112 for now.

To-do:
- [x] Initial implementation
- [x] Add LibreTranslate support
- [x] Per-backend settings and language lists
- [x] Dependency updates
- [x] Show a loading view on translator init
- [x] Support for switching LibreTranslate instances
- [x] Error handling for Translator initialization failure and support for retrying initialization
- [x] Basic instance URL validation support
- [x] Decide on a design for when URL validation fails
- [x] Some cleanup
- [x] Update Search Provider

### DIFF
--- a/data/resources/preferences.ui
+++ b/data/resources/preferences.ui
@@ -69,10 +69,98 @@
                 <property name="subtitle" translatable="yes">Give URL for a specific instance</property>
                 <property name="visible">True</property>
                 <child>
-                  <object class="GtkEntry" id="backend_instance">
+                  <object class="GtkStack" id="backend_instance_stack">
                     <property name="visible">True</property>
-                    <property name="can_focus">True</property>
-                    <property name="valign">center</property>
+                    <property name="can-focus">False</property>
+                    <property name="transition-type">crossfade</property>
+                    <child>
+                      <object class="GtkBox">
+                        <property name="visible">True</property>
+                        <property name="can-focus">False</property>
+                        <property name="spacing">6</property>
+                        <property name="valign">center</property>
+                        <property name="halign">end</property>
+                        <child>
+                          <object class="GtkLabel" id="backend_instance_label">
+                            <property name="visible">True</property>
+                            <property name="can_focus">True</property>
+                          </object>
+                        </child>
+                        <child>
+                          <object class="GtkButton" id="backend_instance_edit">
+                            <property name="visible">True</property>
+                            <property name="can-focus">True</property>
+                            <property name="receives-default">True</property>
+                            <property name="tooltip-text" translatable="yes">Edit</property>
+                            <style>
+                              <class name="image-button"/>
+                            </style>
+                            <child>
+                              <object class="GtkImage">
+                                <property name="visible">True</property>
+                                <property name="can-focus">False</property>
+                                <property name="icon-name">document-edit-symbolic</property>
+                              </object>
+                            </child>
+                          </object>
+                        </child>
+                      </object>
+                      <packing>
+                        <property name="name">view</property>
+                      </packing>
+                    </child>
+                    <child>
+                      <object class="GtkBox">
+                        <property name="visible">True</property>
+                        <property name="can-focus">False</property>
+                        <property name="valign">center</property>
+                        <style>
+                          <class name="linked"/>
+                        </style>
+                        <child>
+                          <object class="GtkEntry" id="backend_instance">
+                            <property name="visible">True</property>
+                            <property name="can_focus">True</property>
+                          </object>
+                        </child>
+                        <child>
+                          <object class="GtkButton" id="backend_instance_reset">
+                            <property name="visible">True</property>
+                            <property name="can-focus">True</property>
+                            <property name="receives-default">True</property>
+                            <property name="tooltip-text" translatable="yes">Reset</property>
+                            <child>
+                              <object class="GtkImage">
+                                <property name="visible">True</property>
+                                <property name="can-focus">False</property>
+                                <property name="icon-name">view-refresh-symbolic</property>
+                              </object>
+                            </child>
+                          </object>
+                        </child>
+                        <child>
+                          <object class="GtkButton" id="backend_instance_save">
+                            <property name="visible">True</property>
+                            <property name="can-focus">True</property>
+                            <property name="receives-default">True</property>
+                            <property name="tooltip-text" translatable="yes">Save</property>
+                            <style>
+                              <class name="suggested-action"/>
+                            </style>
+                            <child>
+                              <object class="GtkImage">
+                                <property name="visible">True</property>
+                                <property name="can-focus">False</property>
+                                <property name="icon-name">emblem-ok-symbolic</property>
+                              </object>
+                            </child>
+                          </object>
+                        </child>
+                      </object>
+                      <packing>
+                        <property name="name">edit</property>
+                      </packing>
+                    </child>
                   </object>
                 </child>
               </object>
--- a/data/resources/preferences.ui
+++ b/data/resources/preferences.ui
@@ -63,6 +63,20 @@
                 <property name="visible">True</property>
               </object>
             </child>
+            <child>
+              <object class="HdyActionRow" id="backend_instance_row">
+                <property name="title" translatable="yes">Specify translator instance</property>
+                <property name="subtitle" translatable="yes">Give URL for a specific instance</property>
+                <property name="visible">True</property>
+                <child>
+                  <object class="GtkEntry" id="backend_instance">
+                    <property name="visible">True</property>
+                    <property name="can_focus">True</property>
+                    <property name="valign">center</property>
+                  </object>
+                </child>
+              </object>
+            </child>
           </object>
         </child>
         <child>
--- a/data/resources/preferences.ui
+++ b/data/resources/preferences.ui
@@ -58,15 +58,15 @@
             </child>
             <child>
               <object class="HdyComboRow" id="backend">
-                <property name="title" translatable="yes">Select Translator</property>
-                <property name="subtitle" translatable="yes">Choose between the available translation backends.</property>
+                <property name="title" translatable="yes">Translator</property>
+                <property name="subtitle" translatable="yes">Choose from the available translation services.</property>
                 <property name="visible">True</property>
               </object>
             </child>
             <child>
               <object class="HdyActionRow" id="backend_instance_row">
-                <property name="title" translatable="yes">Specify translator instance</property>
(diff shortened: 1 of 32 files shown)


rafaelmardojai:
I think we need a `Translator` class witch the UI talks to, and this class manages the history and calls the selected provider.

>  Do we simply clear history each time backend is switched or do we maintain separate histories for each backend?

For now we can clear it each time backend is switched.

> On a similar note, how do we handle saving language preferences? Should we just use some defaults that are available in all services and then reset every time backend is switched or should we maintain separate defaults and separate saved languages?

Defaults and recent languages should be saved per backend I think.

> What do we do for the UI for switching? Just an HdyComboRow? How badly or how well does it handle long lists? (if we end up implementing too many backends)

I don't think we are going to have many backends for now. I don't know about the performance of `HdyComboRow`, but soon we will port to GTK4/libadwaita and I hope things will be better.

>  How many backends should be available? How many is too many?

For now, looks like `libretranslate.com / Argos Translate` is the only backend we can implement. We currently don't know if #47 or #114 are viable.

fushinari:
> I think we need a Translator class witch the UI talks to, and this class manages the history and calls the selected provider.

My current planned approach is that every specific translation service's class will be inherited from a `BaseTranslator` class and they will be called based on a gsettings value at initialization. The results of `translate()` should either be an instance of the `Translation` class or something that inherits from it.

The history will be stored in the class instance that's initiated, which will mean that the history clearing takes place automatically.

The basic idea is to make everything return data similar to the py-googletrans API. For now, we're importing `GTranslator` directly but my plan is to have everything be automatically listed, called, etc. Kinda like the ["servers" implementation in Komikku](https://gitlab.com/valos/Komikku/-/tree/master/komikku/servers).

Is the plan fine as is or should we make changes?

rafaelmardojai:
> Is the plan fine as is or should we make changes?

Looks good to me! Go ahead.

fushinari:
There's a lot more work to do. Adding and updating dependencies (make httpx explicit, add langdetect), making the translators switchable (right now, I'm hardcoding for testing), fixing the visibility of the voice button when switching translators,  etc. But I'm taking a break from this for maybe two days. @rafaelmardojai if you want to commit in the meantime, go ahead. If not, no worries, I'll be back soon anyway. :)

fushinari:
I'm too tired to update flatpak dependencies now, but it's finally in an actually working state. Just barely but it's working. It's now possible to switch between LibreTranslate (the astian.org instance) and Google Translate.

fushinari:
Flatpak dependencies have been updated, so anyone who wants to check this out can try the latest CI build and it'll hopefully work well.

@rafaelmardojai could you check for any glaring bugs?

I'll be taking a long break now (about a month) due to exams.

fushinari:
To-do:
- [ ] Per-backend Settings
- [ ] Some cleanup

rafaelmardojai:
I'm going to test and try to finish the remaining tasks if I find the time.

rafaelmardojai:
Tested, and To-do list updated .

Now we have per-backend settings and a loading view for backends init. 

Please test when you have time.

fushinari:
I tested and it works really well! So, I've added in support for changing the LibreTranslate instance (which is a must-have according to the author of LibreTranslate) but I introduced a bug along the way where an invalid address will get us stuck on the loading page because connection fails. So, there are a few ways to fix this:

- Check validity of instance address before saving using a test button that will check, handle errors and fall back to defaults or save the new address depending on if it's valid.
- Just do error handling and send notifications to the user saying it's misconfigured

I think the first method might be a cleaner one just because it lets the user know beforehand if something is wrong. Gonna sleep for now and I really need to get started with studies :sweat_smile: (I've kinda just been slacking away still)

rafaelmardojai:
I have made some progress. Now we have an initial "backend loading error view", so we can handle any error. And also added a new work flow to edit the backend instance that will allows us to validate the URI before saving, also it will reload the backend every time you change the instance because it could have different languages.

Please test, as a break of the hard study 😅.

rafaelmardojai:
How do you think we should validate the address? Maybe the backend should have a `validate_instance_url` function?

fushinari:
> Please test, as a break of the hard study :sweat_smile:.

Thank you! It's been hard... I tested it and it works really well from what I see!

> How do you think we should validate the address? Maybe the backend should have a `validate_instance_url` function?

That is what I had in mind too, but then again, how would we check if an instance is valid without calling call the functions? In the case of LibreTranslate, maybe we can just check the [frontend settings](https://libretranslate.com/docs/#/frontend/get_frontend_settings).

There's one more issue with supporting LibreTranslate instances. Character limits can be different between instances. The LibreTranslate.com instance limits it to 5000 characters but the Astian instance allows unlimited characters. I was thinking of allowing character limits to be per-backend, but then again, maybe it will lead to abuse and a hard in-app limit is good to have?

rafaelmardojai:
I didn't see your commits and I had some trouble 😂. I have added an initial verification code.

It's 2AM here, so probably I don't know what I'm doing. I'm going to sleep now.

data/resources/preferences.ui
```diff
@@ -56,6 +56,108 @@
...
+                          </object>
+                        </child>
+                        <child>
+                          <object class="GtkButton" id="backend_instance_edit">
+                            <property name="visible">True</property>
+                            <property name="can-focus">True</property>
+                            <property name="receives-default">True</property>
+                            <property name="tooltip-text" translatable="yes">Edit</property>
+                            <style>
+                              <class name="image-button"/>
```
fushinari, line 96: Do we want `image-button`?

With `image-button`:
image

Without it:
image

Also, using `set_image` apparently also sets `image-button`. And using `image-button` with `suggested-action` makes the icon invisible. At first, I just thought "GTK Bug :/" and worked around it, but it's apparently intended behavior. At least one part of it is.

fushinari: This is the bug I'm talking about:
image

fushinari:
> I didn't see your commits and I had some trouble :joy:. 

Ah, sorry for that, just do `git pull --rebase` each time before pushing and we'll be fine I think. :sweat_smile: 

I implemented a sort of error popup. But design needs some fixing. Actually, I'm not even sure that using a popover is good design. I'm quite bad at design in general, so I'm going to leave that up to you. :sweat_smile: 

Current form:


Possible alternatives:

fushinari:
I've also moved the to-do list to the original post and added all the stuff we've already completed just to keep track of all of it. :)

rafaelmardojai:
> I implemented a sort of error popup. But design needs some fixing. Actually, I'm not even sure that using a popover is good design. I'm quite bad at design in general, so I'm going to leave that up to you. 

I prefer the error borders only in the entry.

fushinari:
> I prefer the error borders only in the entry.

Yep, done!

rafaelmardojai:
Nice. 

Now comes the fun part of updating the search provider... 😂

fushinari:
Oh yeah... Before that, we'd also have to disable the backend switcher right?

rafaelmardojai:
> Before that, we'd also have to disable the backend switcher right?

What do you mean?

fushinari:
During validation, don't we want the backend switcher to also be insensitive?

fushinari:
2928a51 is what I meant

fushinari:
Wow... what do we do for the search provider?

data/resources/preferences.ui
fushinari: @rafaelmardojai what do you think about this?

rafaelmardojai:
> Wow... what do we do for the search provider?

Mmm, probably the same as now but checking the backend setting and importing the dialect translators module.

fushinari:
Ah, for some reason, I assumed that the dialect package would be inaccessible from the search provider.

data/resources/preferences.ui
fushinari: Fixed in e508b90

fushinari:
I think it's finally ready. Bit more testing and I think we can merge.

fushinari:
@rafaelmardojai do you have any pending changes? I want to rebase if possible

rafaelmardojai:
No sorry, feel free to rebase. 

We also need to add `preferences.py` to `POTFILES`

fushinari:
Oof... that was not the result I intended...

fushinari:
@rafaelmardojai If I rebase, all your commits are going to show up as "Authored by you and committed by me". Should I just merge instead?

rafaelmardojai:
Yeah no problem

fushinari:
@rafaelmardojai I'm having a very stable experience with this so far.

What do you think we should do next? Call on translators, update and release? Or should we wait, test a bit more, and then release? Also, did you get to test the search provider? Oh, also, do you think the preferences' descriptions right now are okay?

So many questions :laughing: 

On another note, I'm really glad we've finally wrapped up work on the longest standing request so far! LibreTranslate support is complete even with support for using any instance and the framework is there for more translation services! I'm really happy with this!

rafaelmardojai:
I want to put a label in the loading view, then we can call translators for `1.2.0`?

Preferences' descriptions  looks good to me.

fushinari:
Yep, sounds good, go on. :)

TheOPtimal:
I wanted to mention Argos Translate, which LibreTranslate actually uses on the backend. It runs completely offline, and is available as a [python library.](https://github.com/argosopentech/argos-translate)

fushinari:
@TheOPtimal I did consider that but then, I saw that it requires trained models to be manually downloaded by the user and other manual intervention. I also realized that a user of Argos Translate could easily setup a local instance of LibreTranslate if he wishes to. So, to make it as much of an out-of-the-box experience as possible while still giving the user enough power, I ended up deciding on this compromise.

fossdd:
Awesome change!

PR #122, opened 2021-02-09
app icon: new background glyph, better contrast, better shading

bertob:
The current icon has a few issues in my view:
- the glyph in the background is not very geometric, so it's off the pixel grid
- the flat shading for a bent surface looks off
- the D has little contrast
- the color purple/gray combination feels flat, and not very interesting

Here's an attempt to fix the above:

image

This also adds a `.Source` file, since there currently isn't one in the repo.

cc @jimmac


rafaelmardojai:
Thanks for taking the time, looks like a good improvement!  

How looks this with purple? Just curious.

bertob:
I tried purple, but I didn't really like the palette purple in combination with the light gray:

image

rafaelmardojai:
Yeah, in this case the green looks definitely better.
I'm fine with the change, @fushinari what do you think?

fushinari:
I like it.

To be honest, I like the purple one more than the green one, but both look great, so I'm fine with either. And if the general consensus is that green looks better, let's go with that.

fushinari:
I'm ready to merge if @jimmac also approves :)

jimmac:
I approve :)

fushinari:
Thank you for your work! Always great to see amazing designers like you two investing time into the app ecosystem as well.

PR #134, opened 2021-02-20
Update German Translation

Etamuk:
Updated german translation for upcoming version of Dialect


fushinari:
Thank you for your work!

PR #136, opened 2021-03-07
Update Persian translation

musicsd:


rafaelmardojai:
Thanks a lot for taking the time!

Issue #139, opened 2021-03-26
Add options to configure define default input/output language


zeten30:
I think it will be nice to have an option to define input & output language via command line options and in settings.

Example:
```flatpak run com.github.gi_lom.dialect -t "How are you?" -f english -o french```

+ new options in settings.

Otherwise, it tends to use 'auto' input language and it forces user to pick one manually.
![Screenshot from 2021-03-26 10-46-37](https://user-images.githubusercontent.com/1825873/112614064-1d34da00-8e21-11eb-89d8-b1507cfd16f3.png)

rafaelmardojai:
Well, the `-t` parameter was added specifically for the search provider witch always translate from auto to the latest translated language. Do you have any use case to run Dialect from the terminal?

What do you mean with "new options in settings"? Dialect saves your latest source and destination used languages, but it defaults to `Auto` as source language because that way you don't need to select manually the language every time, if you translate from multiple languages. So do you want a preference to select with what languages Dialect should always init?

> Otherwise, it tends to use 'auto' input language and it forces user to pick one manually.

Why would you need to pick the language manually if it will detect it?

zeten30:
Hi, thanks for quick response.

the problem is that 'auto' doesn't work. It does not translate a string until I select a language (visible on screenshot).

I have a keyboard shortcut that runs dialect and translates text in clipboard.

I think that ability to define from/to language will be handy.

zeten30:
@fushinari TYVM:) I will. That will be nice. 

I still think that option to define fixed lang will be nice.

rafaelmardojai:
We can start thinking in adding lang parameters to the command line, it should check if the lang codes are available in the backend or otherwise use the default `Auto -> Latest used`.

rafaelmardojai:
About adding preferences to define fixed languages, I'm not sure, we will need separate preferences for each backend. Maybe we can add a preference to use the latest used source language instead of Auto on launch.

fushinari:
> Maybe we can add a preference to use the latest used source language instead of Auto on launch.

That sounds like a good option to have. Especially because language detection can sometimes be horribly off in LibreTranslate.

rafaelmardojai:
@zeten30 Now we have a preference to disable "Auto" as the default language on every launch. We also have added command line options to define languages codes `--src or -s` for the source language and `--dest or -d` for the destination language.

[...]

PR #140, opened 2021-03-26
Fix launching with "-t" parameter

fushinari:
Fixes an issue mentioned in #139 where launching the application with "-t" would not translate the input terms. Don't know if this is good enough to consider #139 to be fixed.

@zeten30 could you test?


fushinari:
Seems to work as expected in my testing.

rafaelmardojai (approved):
I don't have time to test, but the code looks good.

zeten30:
Thank you! https://github.com/gi-lom/dialect/commit/e3e0ea118f3c7b8279278654f72988afce0443c9 works for me;)

Issue #142, opened 2021-03-30
[Feature Request] add 'listen' button for source text


iman-salmani:
Hi
this feature useful for translate **documents** and learn **speak** source language 
thanks for this app

fushinari:
This is an excellent suggestion! I'll look into implementing this.

PR #145, opened 2021-04-02
Text to Speech improvements 

rafaelmardojai:


rafaelmardojai:
Should I write the settings logic, although we are not going to expose in the UI for now? 

Something like `0 = Disabled` `1 = Google TTS`, and default to `1`.

Maybe we can expose it like a switch to enable/disable TTS in the preferences window, what do you say?

rafaelmardojai:
I have added the preference to enable/disable TTS. I just need to update the window code to follow it.

src/tts/gtts.py
```diff
@@ -0,0 +1,27 @@
...
+    def download_voice(self, text, language, file):
+        try:
+            tts = gTTS(text, lang=language, lang_check=False)
+            tts.write_to_fp(file)
+            file.seek(0)
+
+        except Exception as exc:
+            raise TextToSpeechError(exc) from exc
+
+
```
fushinari, line 27: Extremely minor nitpick, one newline.

Haven't tested yet, but the code looks fantastic so far!

rafaelmardojai: Fixed!

rafaelmardojai:
Looks like is working fine so far, but maybe I'm forgetting something. Only testing will say 😂

fushinari:
How about also doing #142 ?

rafaelmardojai:
> How about also doing #142 ?

I don't have time to do more today, feel free to push if you have time to do it.

fushinari:
No worries, I'll do it tomorrow in another PR. Going to test and merge this.

PR #149, opened 2021-04-04
Preferences: Add option to enable/disable auto translation on launch

### LINKED ISSUE #139: Add options to configure define default input/output language

rafaelmardojai:
Fixes #139

### DIFF
--- a/data/resources/preferences.ui
+++ b/data/resources/preferences.ui
@@ -57,6 +57,21 @@
               </object>
             </child>
             <child>
+              <object class="HdyActionRow">
+                <property name="title" translatable="yes">Auto translation on startup</property>
+                <property name="subtitle" translatable="yes"></property>
+                <property name="visible">True</property>
+                <property name="activatable_widget">auto_trans</property>
+                <child>
+                  <object class="GtkSwitch" id="auto_trans">
+                    <property name="visible">True</property>
+                    <property name="can_focus">True</property>
+                    <property name="valign">center</property>
+                  </object>
+                </child>
+              </object>
+            </child>
+            <child>
               <object class="HdyComboRow" id="backend">
                 <property name="title" translatable="yes">Translator</property>
                 <property name="subtitle" translatable="yes">Choose from the available translation services.</property>
--- a/data/resources/preferences.ui
+++ b/data/resources/preferences.ui
@@ -58,7 +58,8 @@
             </child>
             <child>
               <object class="HdyActionRow">
-                <property name="title" translatable="yes">Use "Auto" as default language on launch</property>
+                <property name="title" translatable="yes">Default to Auto</property>
+                <property name="subtitle" translatable="yes">Use "Auto" as the default language on every launch</property>
                 <property name="visible">True</property>
                 <property name="activatable_widget">auto_trans</property>
                 <child>
--- a/data/resources/preferences.ui
+++ b/data/resources/preferences.ui
@@ -58,8 +58,7 @@
             </child>
             <child>
               <object class="HdyActionRow">
-                <property name="title" translatable="yes">Auto translation on startup</property>
-                <property name="subtitle" translatable="yes"></property>
+                <property name="title" translatable="yes">Use "Auto" as default language on launch</property>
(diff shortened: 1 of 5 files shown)


data/resources/preferences.ui
```diff
@@ -56,6 +56,21 @@
                 <property name="visible">True</property>
               </object>
             </child>
+            <child>
+              <object class="HdyActionRow">
+                <property name="title" translatable="yes">Auto translation on startup</property>
```
rafaelmardojai, line 61: Any ideas for the text here?

src/preferences.py
```diff
@@ -72,6 +73,8 @@ def setup(self):
                            Gio.SettingsBindFlags.DEFAULT)
         self.settings.bind('translate-accel', self.translate_accel,
                            'selected-index', Gio.SettingsBindFlags.DEFAULT)
+        self.settings.bind('auto-trans', self.auto_trans, 'active',
```
rafaelmardojai, line 76: I have a crash sometimes when switching the preference

rafaelmardojai:
I have added command line options to define languages, but haven't time to test it in real world.

src/preferences.py
fushinari: Yeah, I reproduced it twice too but now that I'm ready to debug it, it just won't crash.

src/main.py
```diff
@@ -46,19 +51,29 @@ def do_activate(self):
...
+                langs=self.launch_langs,
                 settings=self.settings
             )
         self.window.present()
 
     def do_command_line(self, command_line):
         options = command_line.get_options_dict()
         options = options.end().unpack()
+        text = ''
+        langs = [None, None]
```
fushinari, line 63: Probably should use a dict here for readability.

data/resources/preferences.ui
fushinari: `Use "Auto" as default language on launch` maybe?

src/main.py
rafaelmardojai: Done!

data/resources/preferences.ui
rafaelmardojai: That sounds beter, at least for now.

rafaelmardojai: That sounds better, at least for now.

src/preferences.py
rafaelmardojai: It seems to only happen when I open the preference window a second time.

rafaelmardojai: And I can reproduce it with all the others switches.

fushinari: Looks like a GTK bug or something, I don't think we're doing anything unconventional here.

rafaelmardojai: Just for the record, it only happens with `Gio.Settings.bind`, when you save the setting manually works fine.

rafaelmardojai: Calling `Gio.Settings.unbind` when `DialectPreferencesWindow` is destroyed fixed the issue. 

So looks like the preference controls are not being finalized when `DialectPreferencesWindow` is destroyed.

From  `Gio.Settings.unbind`  docs:

> Note that bindings are automatically removed when the object is finalized, so it is rarely necessary to call this function.

rafaelmardojai: Funny thing, it only happens when binding `GtkSwitch` but not with `HdyComboRow`.

fushinari: So, GtkSwitch is not finalizing then?

> Just for the record, it only happens with Gio.Settings.bind, when you save the setting manually it works fine.

Yes, I'd seen this too and was rewriting it to be completely manual (even though that's not ideal). Glad to see there's another way.

fushinari: Can confirm. The issue has been fixed. Still kinda weird that we have to manually unbind though...

PR #150, opened 2021-04-05
Dynamic modules

rafaelmardojai:
This initial code should allow us to load dynamically the translation backends.

Currently it only will save us some time when adding a new backend (which doesn't a common task) so I don't know if we will make it reach `main`. But it will help a bit when playing around with experimental backends.

ToDo:
- [x] Update `backend` gsetting to store a string
- [x] Don't convert `TRANSLATORS` to a list, and access it using the new gsetting
- [x] Save the recent langs with a `a{sas}` type, so we don't have to create new gsettings for each backend:
     Example: ```{"google": ["en", "fr", "es", "de"]}```
    And make each backend load it's default values to it if they key doesn't exist?
- [x] Do something to migrate the old gsettings values
- [x] Make TTS use the same system

### DIFF
--- a/data/com.github.gi_lom.dialect.gschema.xml
+++ b/data/com.github.gi_lom.dialect.gschema.xml
@@ -27,9 +27,8 @@
     <key type="s" name="backend-settings">
         <default>"{}"</default>
     </key>
-    <key type="i" name="tts">
-        <!-- 0 = Disabled | 1 = Google TTS -->
-        <default>1</default>
+    <key type="s" name="tts-name">
+        <default>"google"</default>
     </key>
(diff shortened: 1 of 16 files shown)


fushinari:
Next, we need to change the tts to use the same mechanisms and we need a LOT of testing. I've done way too little testing in fact. Haven't even tested what would happen if i deleted a backend yet.

fushinari:
Just tested, removing gtrans.py does not affect the application's function anymore. So yeah, implementing the exact same system for TTS should suffice to complete the request.

fushinari (approved):

PR #154, opened 2021-04-24
window: Focus text view at startup

liferooter:
It will be better to focus text view at startup. You can just open app and type.


rafaelmardojai:
The text view will gain focus as you type, but adding this don't harm.

fushinari (approved):
This change was already made in the GTK4 branch (See https://github.com/dialect-app/dialect/commit/356a7ff134abea83bccc017119bc5e9a635ff64a) . In the case of the main branch, it has no real impact as @rafaelmardojai mentioned. But having visual focus as well is still good behavior. So it's still appreciated :)

Issue #157, opened 2021-04-27
Update translations for 1.3.0


fushinari:
Dialect 1.3.0 is almost ready. As always, we would be grateful if you're able to find time to update the translations. :smile: 

Thank you to all translators who have stuck with us through each release and thank you for allowing users of your specific languages to use the application without having to know a non-native language! I hope you realize that your efforts are recognized and very much appreciated. :smile: 

- [x] German @Etamuk
- [x] Spanish @rafaelmardojai 
- [x] Persian @musicsd
- [x] Hungarian @meskobalazs
- [x] Indonesian @cho2
- [x] Italian @albanobattistella
- [x] Dutch @Vistaus
- [x] Brazilian Portuguese @xfgusta
- [x] Russian @nexeleriot
- [x] Swedish @eson57
- [x] Turkish @v30xy
- [x] Croatian @milotype 
- [x] Czech @pervoj 

We have no set release date, we'll just release when everything is in place. So no need to sweat over it. :smile:

pervoj:
Done #159

Vistaus:
Done! 🙂 #161

Etamuk:
German translation is also done :-)

meskobalazs:
Done at last :)

PR #158, opened 2021-04-27
Updated Turkish Translations

v30xy:
#157

### DIFF
--- a/po/tr.po
+++ b/po/tr.po
@@ -7,7 +7,7 @@ msgstr ""
 "Project-Id-Version: dialect\n"
 "Report-Msgid-Bugs-To: \n"
 "POT-Creation-Date: 2021-04-27 08:53-0500\n"
-"PO-Revision-Date: 2021-04-27 18:24+0300\n"
+"PO-Revision-Date: 2021-04-27 18:42+0300\n"
 "Last-Translator: Bedirhan Tuhana GAYRETLİ <v30xythedev@gmail.com>\n"
 "Language-Team: Turkish <LL@li.org>\n"
 "Language: tr\n"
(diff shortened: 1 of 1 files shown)


po/tr.po
```diff
@@ -132,12 +131,13 @@ msgid "Unselected choice will be used for line break."
 msgstr "Seçilmemiş seçim satır sonu için kullanılacaktır."
 
 #: data/resources/preferences.ui:61
+#, fuzzy
 msgid "Default to Auto"
-msgstr ""
+msgstr "Otomatiği Varsayılan Yap"
```
v30xy, line 136: I'm not sure about here but, its correct i think.

po/tr.po
```diff
@@ -71,10 +71,9 @@ msgstr "Pano butonları"
 
 #: data/com.github.gi_lom.dialect.metainfo.xml.in:26
 msgid "The Dialect Authors"
-msgstr ""
+msgstr "Dialect Yaratıcıları"
 
 #: data/resources/about.ui:11
-#, fuzzy
 msgid "Copyright 2020-2021 The Dialect Authors"
 msgstr "Telif Hakkı 2020 gi-lom"
```
fushinari, line 78: "Telif Hakkı 2020 gi-lom" should probably be "Telif Hakkı 2020-2021 Dialect Yaratıcıları" right?

po/tr.po
rafaelmardojai: It's the title for `Use "Auto" as the default language on every launch`, so feel free to put the title that better represents the description.

fushinari: It should just convey that enabling it will mean that the language selected for the text to translate will be "Auto" on every launch. If the user disables it, the language will be set to the language chosen on last launch.

fushinari: Translating this with Google Translate, I think this is enough :smile:  But then again, I don't know Turkish :sweat_smile:

po/tr.po
v30xy: Oops...

po/tr.po
v30xy: Okay. Thank you all.

fushinari:
Is it ready to merge now?

v30xy:
Yes.

fushinari:
Thank you for the quick work!

PR #163, opened 2021-04-27
Update de.po

Etamuk:
Updated translation for Dialect version 1.3.0


rafaelmardojai:
Thanks a lot! Very appreciated.

PR #167, opened 2021-04-29
Update README.md

milotype:
- add some information about translations of the app and invite users to translate the app


fushinari:
Thank you! I hadn't realized this was needed, I'll be improving upon this soon. :)

PR #174, opened 2021-05-17
Add Portuguese translation

JulianoSCbah:


rafaelmardojai:
Thanks for taking the time!

Could you add yourself to `TRANSLATORS.md` so we can easily notify you to update the translation before releasing new versions?

JulianoSCbah:
Yes, I can.

rafaelmardojai:
Thanks!

PR #175, opened 2021-05-17
Make app name in About window translatable

pervoj:
If you made `about.ui` with Glade, you must mark the application name as translatable manually from a text editor. Otherwise, the translated application name will not be reflected here.

### DIFF
--- a/data/resources/about.ui
+++ b/data/resources/about.ui
@@ -7,7 +7,7 @@
     <property name="modal">True</property>
     <property name="destroy_with_parent">True</property>
     <property name="type_hint">dialog</property>
-    <property name="program_name">Dialect</property>
+    <property name="program_name" translatable="yes">Dialect</property>
     <property name="copyright" translatable="yes">Copyright 2020-2021 The Dialect Authors</property>
     <property name="comments" translatable="yes">A translation app for GNOME.</property>
     <property name="website">https://github.com/dialect-app/dialect</property>


fushinari:
Thank you!

Btw, to anyone wondering, the translation of "Dialect" should be a transliteration rather than an actual translation. So, it would be whatever pronounces the same as or similar to "Dialect" in your language.

pervoj:
> Btw, to anyone wondering, the translation of "Dialect" should be a transliteration rather than an actual translation. So, it would be whatever pronounces the same as or similar to "Dialect" in your language.

Yes, "Dialekt" in Czech. :)

fushinari:
Yep, just said it here in case anyone sees this PR and wonders about it :)

rafaelmardojai:
For the record, the app name was translatable from the `.desktop` file and window title but not from the about dialog because it was intended to represent the app public name rather than the transliterated one.

pervoj:
> For the record, the app name was translatable from the `.desktop` file and window title but not from the about dialog because it was intended to represent the app public name rather than the transliterated one.

@rafaelmardojai Yes, I understand, but there is a link to this repo in the about dialog, so everyone can get to this repository and read the original app name here. If you look to any other app, they have translatable name also in the about dialog. For example GNOME Tweaks:
obrazek

And also I think, if the translators will follow the rule written by @fushinari, it will be enough.

fushinari:
@pervoj Yep, taking the same facts into consideration, we've decided to keep the change. :)

rafaelmardojai:
Yeah, no problem about the change, I just wanted to mention why it was like that before.

fushinari:
@pervoj sorry that this happened:

image

I was simply trying the "Revert" button on GitHub because I've never used it before. Didn't expect it to instantly create a branch and commit. :sweat_smile: Have since deleted the branch.

pervoj:
@fushinari @rafaelmardojai OK, thanks :)

PR #177, opened 2021-05-22
Add Hebrew Translation

yarons:
I could also advise using Weblate for translation contribution, should ease the contribution and increase engagement.


fushinari:
Thank you for your work!

> I could also advise using Weblate for translation contribution, should ease the contribution and increase engagement.

Seems really interesting. Thanks for the suggestion! We'll consider it. :smile:

PR #181, opened 2021-05-22
Spelling: Various fixes

comradekingu:


data/resources/window.ui
```diff
@@ -723,7 +723,7 @@
                                     <property name="sensitive" bind-source="switch_btn" bind-property="sensitive">True</property>
                                     <property name="can-focus">True</property>
                                     <property name="receives-default">True</property>
-                                    <property name="tooltip-text" translatable="yes">Switch languages</property>
+                                    <property name="tooltip-text" translatable="yes">⮂</property>
```
rafaelmardojai, line 726: It's a tooltip text and should be descriptive with the action of the button.

rafaelmardojai, line 726: It's a tooltip text and should be descriptive with the action of the button.

rafaelmardojai (changes requested):

data/resources/window.ui
comradekingu, line 726:
```suggestion
                                    <property name="tooltip-text" translatable="yes">Switches the languages</property>
```

data/resources/preferences.ui
```diff
@@ -52,14 +52,14 @@
...
-                <property name="subtitle" translatable="yes">Unselected choice will be used for line break.</property>
+                <property name="subtitle" translatable="yes">The unselected choice will be used for line break.</property>
                 <property name="visible">True</property>
               </object>
             </child>
             <child>
               <object class="HdyActionRow">
                 <property name="title" translatable="yes">Default to Auto</property>
-                <property name="subtitle" translatable="yes">Use "Auto" as the default language on every launch</property>
+                <property name="subtitle" translatable="yes">Use "Auto" as the default language every time the program is started</property>
```
fushinari, line 62: This is too long. It should be 55 characters or less.

data/resources/preferences.ui
```diff
@@ -173,7 +173,7 @@
             <child>
               <object class="HdyActionRow" id="tts_row">
                 <property name="title" translatable="yes">Text to Speech</property>
-                <property name="subtitle" translatable="yes">Use Google for Text to Speech functionality.</property>
+                <property name="subtitle" translatable="yes">Use Google for speech synthesis functionality.</property>
```
fushinari, line 176: Maybe make this "text-to-speech" instead. Text-to-speech is a more common usage than speech synthesis, so I'd rather keep using the term.

data/resources/preferences.ui
```diff
@@ -190,7 +190,7 @@
         <child>
           <object class="HdyPreferencesGroup" id="search_provider">
             <property name="title" translatable="yes">Search Provider</property>
-            <property name="description" translatable="yes">Dialect has a GNOME Search Provider. Enabling or disabling Live Translation here will also enable or disable the GNOME Search Provider. This is to reduce the possibility of API abuse. Also make sure that the feature is enabled in GNOME Settings.</property>
+            <property name="description" translatable="yes">The GNOME search provider in Dialect is turned off when Live translation is to reduce API abuse. Ensure it is also turned on in the GNOME Settings.</property>
```
fushinari, line 193: To avoid confusion, maybe make this "In order to reduce API abuse, Dialect's GNOME search provider is turned off when Live Translation is. Ensure that the search provider is also turned on in GNOME Settings." ?

data/resources/window.ui
```diff
@@ -814,7 +814,7 @@
                             <property name="visible">True</property>
                             <property name="can-focus">False</property>
                             <property name="halign">start</property>
-                            <property name="label" translatable="yes">No network connection detected.</property>
+                            <property name="label" translatable="yes">Connect to the Internet first.</property>
```
fushinari, line 817: Probably better to leave this as is. We don't want to tell the user what to do, we only want to tell the user why the application isn't functioning as it should.

data/resources/preferences.ui
```diff
@@ -81,7 +81,7 @@
             <child>
               <object class="HdyActionRow" id="backend_instance_row">
                 <property name="title" translatable="yes">Translator Instance</property>
-                <property name="subtitle" translatable="yes">Specify URL of translation service.</property>
+                <property name="subtitle" translatable="yes">Enter a translation service URL.</property>
```
fushinari, line 84: Maybe "Enter URL for translation service"? I mean, it has a default fallback value, so "Enter a translation service URL" doesn't seem right for some reason.

data/resources/window.ui
```diff
@@ -723,7 +723,7 @@
                                     <property name="sensitive" bind-source="switch_btn" bind-property="sensitive">True</property>
                                     <property name="can-focus">True</property>
                                     <property name="receives-default">True</property>
-                                    <property name="tooltip-text" translatable="yes">Switch languages</property>
+                                    <property name="tooltip-text" translatable="yes">Switches the languages</property>
```
fushinari, line 726: Tooltips are not for what a button does, it's for a button will do. So, maybe better to leave it be.

fushinari (changes requested):
Thanks for working on this! Strings are often a very tricky matter because they should convey to the user exactly what clicking a button will do but sometimes it's not that simple. It's especially troublesome with neither of us being native English speakers.

Just going to note a few of my concerns with the changes, please comment or make changes accordingly and we can get to merging this.. :smile:

data/resources/preferences.ui
Vistaus: Just a suggestion: maybe just shorten it to ‘Use “Auto” as the default language’? That should be clear enough for the majority of English speakers.

data/resources/preferences.ui
Vistaus: Or, alternatively, ‘Enter the URL of a translation service’?

data/resources/preferences.ui
fushinari: Yep, that should be fine!

data/resources/preferences.ui
fushinari: Maybe. I'm not so sure because all versions feels wrong in my head for some reason :sweat_smile: .

comradekingu: @fushinari Including: "Enter a link to a translation service"?

fushinari: Address feels more appropriate

fushinari: "Enter address of the translation service" was what I originally had in mind, but string length was the issue. Any string that's roughly the same length as the original or less should be fine

data/resources/preferences.ui
comradekingu, line 62:
```suggestion
                <property name="subtitle" translatable="yes">Use "Auto" as the default language</property>
```

data/resources/preferences.ui
comradekingu: I thought this was the description, which doesn't actually do much over saying it 3 times if it is just "text-to-speech".

data/resources/preferences.ui
comradekingu: What exactly does turning on a search provider in the GNOME settings entail?

comradekingu, line 193:
```suggestion
            <property name="description" translatable="yes">To reduce API abuse, the GNOME search provider in Dialect is turned off when Live translation is. Ensure the search provider is also turned on in the GNOME Settings.</property>
```

data/resources/window.ui
comradekingu, line 726:
```suggestion
                                    <property name="tooltip-text" translatable="yes">Switch languages</property>
```

data/resources/window.ui
comradekingu: One is not as obvious a conclusion based on available info.

data/resources/preferences.ui
rafaelmardojai: You aren't going to get results from that provider on the GNOME shell search.

We added that line because depending on some factors (like if the app is installed through flatpak or directly on the host), the search provider can be disabled by default on GNOME search settings.

data/resources/window.ui
fushinari: We want to leave this error as generalized as possible because it could be due to a variety of reasons. The user could actually be connected to the internet for all we know. So it's better to say something like the app not being able to detect a network connection. There may be better wording for it, but we don't want to tell the user to connect to the internet instead.

data/resources/preferences.ui
fushinari: It doesn't matter how many times we say the same thing because the information that we really want to convey in the description is that Google will be used. A bit of repetition for the sake of completeness is fine here imo.

@rafaelmardojai what do you think?

comradekingu: It does matter, because it is redundant.
The original should be text-to-speech, so an error is avoided too.

comradekingu, line 176:
```suggestion
                <property name="subtitle" translatable="yes">Use Google for TTS.</property>
```

data/resources/window.ui
comradekingu:
```suggestion
                            <property name="label" translatable="yes">No network connection detected.</property>
```

comradekingu: That there could be an error in the detection or routing for this specifically is not usually the case. Not a whole lot more correct, but harder to decipher.

data/resources/preferences.ui
comradekingu: Lets revisit this when and if an alternative can be presented within those limits.
The original has a marked noun error, where it should be "the URL", and "a translation service", but that makes it too long. The alternative works.

data/resources/preferences.ui
rafaelmardojai: The new version `Use Google for TTS.` looks good to me.

fushinari: Yeah, that looks good to me too. Could you also fix the title? Like you mentioned, hyphenation is needed (which was why i said "make this text-to-speech" rather than ask to revert but I now realize I wasn't able to say clearly, goes to show how lacking my language skills are :sweat_smile:)

data/resources/preferences.ui
fushinari: Okay, that's good enough for now I guess.

data/resources/preferences.ui
```diff
@@ -173,7 +173,7 @@
             <child>
               <object class="HdyActionRow" id="tts_row">
                 <property name="title" translatable="yes">Text to Speech</property>
```
fushinari, line 175:
```suggestion
                <property name="title" translatable="yes">Text-to-Speech</property>
```

fushinari (approved):
Looks great!

data/resources/preferences.ui
rafaelmardojai: What's wrong with this?

fushinari: Nothing much. Hyphenating it is more correct. It's our bug though. But I was hoping for input when suggesting here

rafaelmardojai: Right. You can commit the suggestion :P

fushinari: Yes, but I wanted to confirm this is what we want :)

PR #187, opened 2021-06-30
Final 1.x changes

rafaelmardojai:
- [x] Add the final accels to the Shortcuts Window.

### DIFF
--- a/data/resources/shortcuts-window.ui
+++ b/data/resources/shortcuts-window.ui
@@ -10,14 +10,7 @@
         <child>
           <object class="GtkShortcutsGroup">
             <property name="visible">True</property>
-            <property name="title" translatable="yes" context="shortcuts window">General</property>
-            <child>
-              <object class="GtkShortcutsShortcut">
-                <property name="visible">True</property>
-                <property name="title" translatable="yes" context="shortcuts window">Preferences</property>
-                <property name="accelerator">&lt;Primary&gt;comma</property>
-              </object>
-            </child>
+            <property name="title" translatable="yes" context="shortcuts window">Translator</property>
             <child>
               <object class="GtkShortcutsShortcut" id="translate_shortcut">
                 <property name="visible">True</property>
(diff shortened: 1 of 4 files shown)


rafaelmardojai:
`Show Pronunciation` and `Clear` are the only actions left. Should we also add accels for those? Suggestions?

rafaelmardojai:
Current accels for further discussion:

- Back and Forward are standard (`Alt+←` `Alt+→`).
- Paste: `Ctrl + Shift + V`
- Copy: `Ctrl + Shift + C`
- Listen src text: `Ctrl + Alt + L`
- Listen dest text: `Ctrl + L`

fushinari:
> Show Pronunciation and Clear are the only actions left. Should we also add accels for those? Suggestions?

`Ctrl + P` and `Ctrl + D` maybe?

> - Paste: Ctrl + Shift + V

No need to document this one i think, because the default focus is on src_text. Only reason I left the paste button there is because I felt it might be convenient for touch users. So maybe keep the shortcut but not document it?

Looks good otherwise

rafaelmardojai:
> Ctrl + P and Ctrl + D maybe?

Done.

Ideas for the Shortcuts Window? Should I add all the actions to the General group? Or how we better structure the actions?

dialect/window.py
```diff
@@ -559,13 +547,13 @@ def on_dest_lang_changed(self, _obj, _param):
     """
     User interface functions
     """
-    def ui_return(self, _button):
+    def ui_return(self, *args):
```
fushinari, line 550: Nitpick. We've been using `_action, _param` so far, so maybe continue using that?

rafaelmardojai: Hmm yeah, I just was thinking that maybe we could need to call the functions directly somewhere else, but in that case probably makes more sense to call the action. patch coming,

data/resources/shortcuts-window.ui
```diff
@@ -32,6 +25,87 @@
...
+              <object class="GtkShortcutsShortcut">
+                <property name="visible">True</property>
+                <property name="title" translatable="yes" context="shortcuts window">Clear text entry</property>
+                <property name="accelerator">&lt;Primary&gt;D</property>
+              </object>
+            </child>
+            <child>
+              <object class="GtkShortcutsShortcut">
+                <property name="visible">True</property>
+                <property name="title" translatable="yes" context="shortcuts window">Copy output text</property>
```
fushinari, line 38: "Copy translation" maybe? I think we want to avoid using terms like "output" in the UI.

data/resources/shortcuts-window.ui
```diff
@@ -32,6 +25,87 @@
                 <property name="accelerator">&lt;Primary&gt;S</property>
               </object>
             </child>
+            <child>
+              <object class="GtkShortcutsShortcut">
+                <property name="visible">True</property>
+                <property name="title" translatable="yes" context="shortcuts window">Clear text entry</property>
```
fushinari, line 31: "Clear source text" maybe?

data/resources/shortcuts-window.ui
```diff
@@ -32,6 +25,87 @@
...
+          </object>
+        </child>
+        <child>
+          <object class="GtkShortcutsGroup">
+            <property name="visible">True</property>
+            <property name="title" translatable="yes" context="shortcuts window">Text-to-Speech</property>
+            <child>
+              <object class="GtkShortcutsShortcut">
+                <property name="visible">True</property>
+                <property name="title" translatable="yes" context="shortcuts window">Listen source text</property>
```
fushinari, line 58: Listen to source text

data/resources/shortcuts-window.ui
```diff
@@ -32,6 +25,87 @@
...
+                <property name="accelerator">&lt;Primary&gt;&lt;shift&gt;L</property>
+              </object>
+            </child>
+            <child>
+              <object class="GtkShortcutsShortcut">
+                <property name="visible">True</property>
+                <property name="title" translatable="yes" context="shortcuts window">Listen output text</property>
```
fushinari, line 65: Listen to translation

fushinari:
I have some cleanup wor locally, separate from the work in this MR. I want to push it, but I also want to test some stuff, so it might be a bit later (college has been annoying :sweat_smile: ). If you want to, you can just merge this, it looks good as is.

Issue #115, opened 2021-01-12
[Feature request] Add proxy setting


Nokia808:
Hi. 

This many be strange request from title ! But it's usefulness is that:
1) unfortunately, there are certain situation in which even such sites like online translations are blocked ! In such cases this feature will be extremely useful ...
2) to improve privacy of users, like utilizing Google via Tor network to avoid Google anti-privacy hanger....

If this feature accepted to be added as future enhancement, then it should allow SOCKS 5 proxy & should allow user to select UDP or TCP or automatic(UDP/TCP) protocol.

Thank you.

fushinari:
Personally, I don't think this is within the scope of the application and any proxy settings should probably be system-wide. What do you think, @rafaelmardojai ?

Nokia808:
@fushinari 
Well, at a time it is more suitable to use system-wide solution. But at other cases it is more suitable to make only certain applications bypass the blockage .......

However, the final decision is up to the developer(s)

fushinari:
Actually, I think I agree with you that it's a good addition. But it adds a layer of complexity that I'm not sure I want. Also, to implement it properly, we would need `py-googletrans` to also support proxy settings, but sadly, it's still marked as a TODO.

Nokia808:
@fushinari 
It is very okay to be in "TODO" list ! Take your time.

Venturub:
I agree with this. And would suggest the user IP to be somehow replaced by another IP who would then be the one directly presenting the translation request to Google Translator. This would further justify the use of Dialect over the Google Translator website through the browser. I do not know the feasibility of this but it would be great... thanks!

rafaelmardojai:
> The python-requests library supports it, thats why it will work for dialect.

Just for the record, Dialect doesn't use `python-requests` but `libsoup`. And it should respect the proxy settings from GNOME Settings.

zer0-x:
Oh

So `python-requests` will be used for Google translate only?
`gTTS` uses request as I saw [here](https://github.com/pndurette/gTTS/blob/3d6cfc9d6da6a83dbfe5e46911f8ac87f19215a7/setup.cfg#L40).

rafaelmardojai:
> So python-requests is used for Google translate only?
[...]

PR #195, opened 2021-09-07
Basic SOCKS proxy support

### LINKED ISSUE #192: Does not work behind a socks proxy server
130969299-c44cefc7-dfb5-426a-93db-5d15de372e8e

fushinari:
Fixes #192 

Based on Lollypop's handling of the same. Unfortunately, this isn't exactly ideal. As discussed in https://github.com/dialect-app/dialect/issues/115#issuecomment-760934645 previously, `py-googletrans` and `httpx` do not properly support proxies. But for now, this should hopefully work.

### DIFF
--- a/dialect/main.py
+++ b/dialect/main.py
@@ -4,8 +4,11 @@
 # SPDX-License-Identifier: GPL-3.0-or-later
 
 # Initial setup
+import socket
+import socks
 import sys
 from gettext import gettext as _
+from os import environ
 
 import gi
 gi.require_version('Gdk', '3.0')
@@ -35,6 +38,8 @@ class Dialect(Gtk.Application):
         self.launch_text = ''
         self.launch_langs = {}
 
+        self.proxy_status = setup_proxy()
+
         # Add command line options
         self.add_main_option('text', b't', GLib.OptionFlags.NONE,
                              GLib.OptionArg.STRING, 'Text to translate', None)
@@ -47,7 +52,7 @@ class Dialect(Gtk.Application):
 
     def do_activate(self):
         self.window = self.props.active_window
-        
+
         if not self.window:
             width, height = Settings.get().window_size
             self.window = DialectWindow(
@@ -60,7 +65,7 @@ class Dialect(Gtk.Application):
                 langs=self.launch_langs
             )
             self.setup_actions_signals()
-            
+
         self.window.present()
 
     def do_command_line(self, command_line):
@@ -80,7 +85,7 @@ class Dialect(Gtk.Application):
             langs['dest'] = options['dest']
 
         if self.window is not None:
-             self.window.translate(text, langs['src'], langs['dest'])
+            self.window.translate(text, langs['src'], langs['dest'])
         else:
             self.launch_text = text
             self.launch_langs = langs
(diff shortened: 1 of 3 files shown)


dialect/main.py
```diff
@@ -218,6 +223,33 @@ def on_quit(self, _action, _param):
...
+        if mode == "manual":
+            # Fetch socks proxy settings
+            socks_settings = Gio.Settings.new("org.gnome.system.proxy.socks")
+            host = socks_settings.get_value("host").get_string()
+            port = socks_settings.get_value("port").get_int32()
+
+            if host != "" and port != 0:
+                # Reset proxy env vars
+                environ["all_proxy"] = ""
+                environ["ALL_PROXY"] = ""
```
fushinari, line 240: I had to do this because something kept reading it from the environment variable and did not allow the app to work. But from my experimentation using NetHogs, the proxy seems to work just fine, so it should be okay.

fushinari:
@bellegarde-c if you find time to, could you test to see if this works as expected in a real-world test? (I could only test with the experimental setup you suggested :sweat_smile:)

dialect/main.py
bellegarde-c: Does not work here (timeout on search). Looks like settings variable is mandatory.

Tried but your right, something manually tries to understand proxy env :(

fushinari: I see :( Too bad. Well... proper proxy support and a bunch of other large changes will be coming eventually, so maybe we will have to put this off until then.

fushinari: For now, converting to draft and will experiment again soon if possible. But unfortunately... it works for me now. So I don't know how to proceed :sweat_smile: .

bellegarde-c: Will try to debug

fushinari: That would be awesome! Thanks!

bellegarde-c:
Ok, tried this code:

```
import httpx
import socks
import socket

socks.set_default_proxy(socks.SOCKS5, "127.0.0.1", 8080)
socket.socket = socks.socksocket

URL = 'http://ifconfig.me/ip'

with httpx.Client() as client:
    resp = client.get(URL)
    print(resp.text)
```

But fails with **httpx.ConnectError: [Errno 97] Address family not supported by protocol**

So until httpx socks support is out https://github.com/encode/httpx/issues/203 and gtrans add support for it, we can't do more IMO.

bellegarde-c:
Hmm, or it's because IPV6 is disabled on my laptop.

fushinari:
> Ok, tried this code (...)

Interesting... This works for me until I enable the proxy in GNOME settings. When I do that, httpx tries to use the environment variable and fails irrespective of it is `socks4` or just `socks`.

rigens:
There is a fairly stable 3rd party SOCKS implementation: [httpx-socks](https://github.com/romis2012/httpx-socks)

fushinari:
> There is a fairly stable 3rd party SOCKS implementation: [httpx-socks](https://github.com/romis2012/httpx-socks)

We use httpx through py-googletrans for Google Translate. So we can't implement this.

fushinari:
Closing this because transition to libsoup should solve this and is being done in #235 .

PR #200, opened 2021-09-16
Languages handing improvements

### LINKED ISSUE #193: Please allow languages to be translatable
Please allow the languages in the drop-downs to be translatable. Please see image below.

kép

What do you think?

Thank you

### LINKED ISSUE #197: Listen button is greyed out when translating from Chinese
I'm using dialect 1.3.0 flatpak version from flathub on Fedora 34. When translating from simplified Chinese the listen button is greyed out. The backend is Google Tanslate.

Screenshot from 2021-09-15 00-21-36

rafaelmardojai:
- Provide localized languages name dictionary
- Standardize lang codes with aa-AA format
- Provide helper func to get pretty + localized lang name or return the
lang code if no match is found
- Update translators to only provide a lang codes list in the
`languages` variable

TODO:

- [x] Add migration code for saved recent used src/dest langs in our `Settings` object.
Update `zh-cn` and `zh-tw` to `zh-CN` and `zh-TW`.

Fixes #197.
Fixes #193.


fushinari:
This will fix #197 and #193 .

fushinari:
In the `po` repository, there is now a little language names puller script that takes localized language names from Google Translate. I'm keeping it there as a tool for future translators.

I actually messed up just a bit by pushing into `main` on `po`. Shouldn't really be a problem to anyone except the ones running on `main`.

I went with the approach in a8afdf6 because adding migration code would need a way to tell apart between older configuration and newer configuration, which we don't have. a8afdf6 also fixes a long standing bug with how the recent languages work, i.e. it now sets an exact limit of 4 recent languages to be saved.

fushinari (approved):

Issue #151, opened 2021-04-09
Google Translate backend only translates to English


miguelangelnieto:
When choosing Google Translate, it only translates to English. See the following screenshot:

![Screenshot from 2021-04-09 20-53-54](https://user-images.githubusercontent.com/810868/114227893-ea88f680-9975-11eb-907b-2ba5022b25e1.png)

Same happens if I choose Spanish to Russian.

I am using Dialect 1.2.0, Flatpak, on Fedora 34.

LyesSaadi:
Happened to me sometimes, but I could not reproduce it consistently :/. Seems to be Google Translate-specific since it never happened to me with LibreTranslate... Maybe is it a temporary Google API ban? Or a googletrans error?

fushinari:
Yeah, my guesses are the same as @LyesSaadi . Sadly nothing we can do about it for now I think.

miguelangelnieto:
Hi, thank you! Is there anything I can do to help in the diagnosis? Any debug flag or anyway to trace the API?

fushinari:
Sorry but the issue is not in Dialect's code, it's in code we depend on. So there's nothing any of us can do about it right now, i.e. while we still depend on said code. I'd still like info if possible though :)

Are you able to consistently reproduce this? If not, how often does it happen? Are you able to use Google Translate normally from your browser? Is the issue specific to some language pairs?

miguelangelnieto:
Yes, I can reproduce it. It happens always, since the first moment after the installation. I choose to translate Spanish to any other language, and it always gets translated to English. Funny thing, if I click the button to get the audio, it uses the language accent. For example, if I translate "Hola" from Spanish to Russian, I will get "Hello", but the audio will be played with strong Russian accent.

In the browser it works without problems.

LyesSaadi:
I misunderstood the issue, I read your issue too quickly, and I now see what you mean. I tried it myself afterward and can confirm it is consistent! I didn't have this issue before, and I don't know what created it exactly?

(I thought you were talking about another bug where after a time translation no longer works and repeatedly give the same answer even when we change the sentence.)

LyesSaadi:
This is maybe a dialect issue? While dialect refuses to translate, here is what googletrans 3.1.0a0 (so using the legacy method and the one in the flatpak) outputs:

```
lyes in dialect on  main [?] 
➜ translate "Bonjour, mon ami !"
[fr] Bonjour, mon ami !
    ->
[en] Hello my friend !
[...]

PR #212, opened 2021-10-01
fix(translators/google): propagate api exceptions

sorin-davidoi:
The default is `False` (https://github.com/ssut/py-googletrans/blob/d15c94f176463b2ce6199a42a1c517690366977f/googletrans/constants.py#L186). Shows "Translation failed. Please check for network issues" when the API returns `429 Too Many Requests`. While the message is a bit misleading (there are no network errors), it's an improvement over the current situation (the input test is shown as the output and no errors are reported to the user).

Closes https://github.com/dialect-app/dialect/issues/151.

### DIFF
--- a/dialect/translators/gtrans.py
+++ b/dialect/translators/gtrans.py
@@ -30,7 +30,7 @@ class Translator(TranslatorBase):
     }
 
     def __init__(self, **kwargs):
-        self._translator = GoogleTranslator()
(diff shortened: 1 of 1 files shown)


rafaelmardojai:
Thanks!

>  While the message is a bit misleading

Yeah true, that can be improved latter. We should split network errors from server ones.

PR #216, opened 2021-10-20
Port to GTK4+Adwaita

rafaelmardojai:
Port to GTK4 and use new libadwaita APIs.

I want to rework the whole `DialectLangSelector`, should I do it here or in a later PR?


dialect/preferences.py
```diff
@@ -49,17 +50,12 @@ def __init__(self, parent, **kwargs):
...
-        model = Gio.ListStore.new(Handy.ValueObject)
-        options = ['Ctrl + Enter', 'Enter']
-        for index, value in enumerate(options):
-            model.insert(index, Handy.ValueObject.new(value))
-        self.translate_accel.bind_name_model(model,
-                                             Handy.ValueObject.dup_string)
 
+        # Show dark mode preference
+        self.style_manager = self.parent.app.get_style_manager()
+        if not self.style_manager.get_system_supports_color_schemes():
```
fushinari, line 56: We could just keep this as a force dark switch instead or as a separate in-app theme switcher in case the user has a separate preference for Dialect than the rest of their system.

dialect/settings.py
```diff
@@ -205,7 +205,7 @@ def get_src_langs(self, backend):
         if settings is not None and settings.get('src-langs'):
             return settings.get('src-langs')
 
-        return TRANSLATORS[backend].src_langs
+        return TRANSLATORS[backend].src_langsP
```
fushinari, line 208: Typo?

dialect/window.py
```diff
@@ -430,36 +469,36 @@ def send_notification(self, text, timeout=5):
...
+            self.src_voice_btn.set_child(self.src_voice_image)
             self.src_voice_spinner.stop()
 
             dest_text = self.dest_buffer.get_text(
                 self.dest_buffer.get_start_iter(),
                 self.dest_buffer.get_end_iter(),
                 True
             )
-            self.app.listen_dest_action.set_enabled(
+            self.lookup_action('listen-dest').set_enabled(
```
fushinari, line 497: Is lookup_action performant?

fushinari:
> I want to rework the whole DialectLangSelector, should I do it here or in a later PR?

In it's current state, it doesn't work at all, so maybe here XD

dialect/lang_selector.py
```diff
@@ -38,7 +38,7 @@ def __init__(self, **kwargs):
 
         self.factory = Gtk.BuilderListItemFactory.new_from_resource(
             None, f'{RES_PATH}/lang-row.ui'
-        );
```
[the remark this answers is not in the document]
fushinari: Lol, I get it 😂

dialect/window.py
rafaelmardojai: Should be, and I haven't noticed any performance regression.

dialect/preferences.py
rafaelmardojai: Right, you think we need a in-app theme switcher?

rafaelmardojai: Do you mean always visible?

fushinari: Yes

rafaelmardojai:
You know if setting our custom shortcuts window with  `Gtk.ApplicationWindow.set_help_overlay` will also work?

rafaelmardojai:
Right, I think we can leave it as is for now.

PR #222, opened 2021-11-07
Port translators settings to GSettings

rafaelmardojai:
TODO:
- [x] Clenup and remove old settings functions
- [x] Update `reset_` functions
- [x] Implement migrations (IMO we should only migrate `backend-name` and `instance-url`. We aren't dealing with sensitive data :P)

### DIFF
--- a/dialect/settings.py
+++ b/dialect/settings.py
@@ -25,8 +25,6 @@ class Settings(Gio.Settings):
 
     def __init__(self):
         Gio.Settings.__init__(self)
-        self.init_translators_settings()
-        self.migrate_legacy()
 
     @staticmethod
     def new():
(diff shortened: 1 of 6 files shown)


dialect/settings.py
```diff
@@ -21,9 +21,12 @@ class Settings(Gio.Settings):
     """
 
     instance = None
+    translator = None
 
     def __init__(self):
         Gio.Settings.__init__(self)
+        self.init_translators_settings()
```
fushinari, line 28: This should be in `new()`

fushinari (commented):
Backend switching does not work reliably right now. Switching back and forth between LibreTranslate and Google will get you stuck on Google at some point.

dialect/settings.py
rafaelmardojai: Yeah, it was in `new()`, then I moved to try a better looking code and it worked, but now I realize the fact that I already had initialized the settings...

fushinari (approved):

PR #225, opened 2021-11-15
AdwToasts for in-app notifications

fushinari:
In return, we lose some functionality and control. We also introduce some warnings and errors that do not effect the functioning of the application.

What we gain:
- Beautifully animated toasts in place of our poorly done in-app notifications system.
- Some cleanup because we get to delete code.

What we lose:
- Icons in notifications (success, failure, warning, etc)
- ~Control over timeout~ AdwToast now has support for setting timeouts.

Other issues:
- Warnings like:
```
(dialect:2): Adwaita-CRITICAL **: Trying to dismiss the toast '5000 characters limit reached!', but it isn't in an AdwToastOverlay yet
```
These do not seem to impact the functioning of the application itself.

### DIFF
--- a/data/resources/window.ui
+++ b/data/resources/window.ui
@@ -350,7 +350,7 @@
                                     <child>
                                       <object class="GtkButton" id="translate_btn">
                                         <property name="label" translatable="yes">Translate</property>
-                                        <property name="sensitive">False</property>
+                                        <property name="action-name">win.translation</property>
                                         <style>
                                           <class name="suggested-action"/>
                                         </style>
--- a/data/resources/window.ui
+++ b/data/resources/window.ui
@@ -212,7 +212,7 @@
                   </object>
                 </child>
                 <child>
-                  <object class="GtkOverlay">
+                  <object class="AdwToastOverlay" id="toast_overlay">
                     <property name="vexpand">True</property>
                     <child>
                       <object class="GtkBox">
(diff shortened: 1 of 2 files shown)


rafaelmardojai:
Doesn't `ToastOverlay` manage the queue itself? With the priorities and that?

fushinari:
> Doesn't `ToastOverlay` manage the queue itself? With the priorities and that? 

It does. But even if you set a higher priority for a new notification, it will only get sent to the beginning of the queue. In our case, like the 5000 characters notification, we don't usually need the earlier notifications at all anymore, so clearing the queue is better.

I'm thinking of improving the queue parameter though...

rafaelmardojai:
Right... Now I get it.

rafaelmardojai:
Should we add actions like "Retry" no network related notifications?

fushinari:
Before we merge this, we want to solve the AdwCritical errors. Ideally, we would get a clear toasts option in AdwToastOverlay or AdwToast.dismiss would ignore the dismiss request if it's already not present in the AdwToastOverlay. This will surely need discussion in libadwaita itself.

Adding a retry button for translation involved switching to using actions for translation. However, this introduces a new styling bug where the "Translate" label is miscolored during the action's transition between enabled and disabled. Looks like a libadwaita or GTK4 bug.

fushinari:
The `GtkWarning`s seem to be gone.

rafaelmardojai:
Hmm, maybe instead of clearing the toasts, we could have some time-out to avoid spamming notifications? Maybe adding ids to the notifications? I don't know, probably it's overcomplicating things... 😅

fushinari:
The errors can't be prevented even if we do that. :(

fushinari:
libadwaita has added support for timeout control now. We can ~ab~use that I guess. Let's see.

fushinari:
I was waiting in case `libadwaita` implemented icons support. But that's not happening it seems (see: [GNOME/libadwaita#334](https://gitlab.gnome.org/GNOME/libadwaita/-/issues/334)).

What do you think about merging this as it is for now? We could make other changes later @rafaelmardojai

rafaelmardojai:
Yep, go ahead.

fushinari:
New Bugs:
- `AdwCritical` errors because of toast queuing issues.
-  Styling bug where the "Translate" label is miscolored during the action's transition between enabled and disabled.

PR #226, opened 2021-11-23
Random polish

Exalm:
Screenshot from 2021-11-23 22-31-54

Screenshot from 2021-11-23 22-32-01

### DIFF
--- a/data/resources/lang-selector.ui
+++ b/data/resources/lang-selector.ui
@@ -15,7 +15,9 @@
           <object class="GtkBox" id="search_box">
             <property name="orientation">vertical</property>
             <child>
-              <object class="GtkSearchEntry" id="search"></object>
+              <object class="GtkSearchEntry" id="search">
+                <property name="placeholder-text" translatable="yes">Search Languages…</property>
+              </object>
             </child>
             <style>
               <class name="search_box"/>
--- a/data/resources/style.css
+++ b/data/resources/style.css
@@ -3,18 +3,14 @@
   padding: 12px;
 }
 
-.translation-side-box {
-  background: @theme_base_color;
-  border: 1px solid @borders;
-  border-radius: 9px;
-}
-
-.translation-side-box :disabled {
-  background: @insensitive_bg_color;
-}
-
 .translation-side-box scrolledwindow {
   border-bottom: 1px solid @borders;
+}
+
+.translation-side-box textview,
+.translation-side-box text {
+  background: none;
+  color: inherit;
 }
 
 .mistakes {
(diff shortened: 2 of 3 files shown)


rafaelmardojai (approved):
LGTM, thanks for taking the time!

fushinari:
@Exalm I think this introduced a bug:

Screenshot from 2021-11-24 01-26-29

Exalm:
Hmm, you can see on my screenshot above I can't reproduce it.

fushinari:
Odd, but it definitely originated in 0fedb76 for me :( Guess I'll look into it tomorrow when I'm not half-asleep :sweat_smile:

Exalm:
Absolutely possible, but how are you building it? What's your gtk version, libadwaita version etc :)

fushinari:
GNOME platform nightly, up-to-date.

Interestingly, removing and adding back the `translation-side-box` style class fixes it temporarily (until the app closes).

rafaelmardojai:
It also works fine for me, I tested it before merging in both light and dark.

Exalm:
> Interestingly, removing and adding back the translation-side-box style class fixes it temporarily (until the app closes).

Wait what. That's weird, not sure how that can even happen...

fushinari:
> Wait what. That's weird, not sure how that can even happen...

Yep it's very weird. What would be good debugging steps in such a case? If adding the style class once again does fix the issue, I guess that means the CSS is fine. So the issue is in GTK?

> It also works fine for me, I tested it before merging in both light and dark.

Seems specific to me then :sweat_smile:

Exalm:
Good question. What happens if you reproduce the bug and then poke at the widget with inspector? CSS Nodes page, I'm interested what `color` there is for the `textview` and `text` nodes.

fushinari:
Even weirder... the color is (255, 255, 255). So it should technically be white but is being rendered as black due to some bug?

It's running on the GNOME nightly runtime, meaning GTK version is 4.5.0, running on Wayland.

Another thing I noticed is that switching the themes from `Adwaita-empty` to `Default` (or any other non-Adwaita theme) and switching back to `Adwaita-empty` also gets rid of the issue.

Exalm:
Right, it's invalidating CSS, similarly.

But yeah, looks like a GTK bug...

fushinari:
I see.. I'll open an issue tomorrow then. Thanks! For everything really XD

fushinari:
@Exalm @rafaelmardojai Could you test for the bug I mentioned by enabling the "Force dark mode" option in Dialect, quitting and then re-launching?

Exalm:
Aha, I can reproduce that now.

Exalm:
While I don't know why exactly it happens I see you're changing color scheme when creating the window. That's fragile, you should do it beforehand.

```python
    def do_startup(self):
        Adw.Application.do_startup(self)

        Gst.init(None)  # Init Gst

        # Load saved dark mode
        if Settings.get().dark_mode:
            self.get_style_manager().set_color_scheme(Adw.ColorScheme.FORCE_DARK)
```

E.g. here.

fushinari:
Thanks for all your help throughout this! Should be fixed now.

Should I still report this in the GTK tracker?

Exalm:
I guess, though I mean you'd need a proper minimal reproducer.

PR #228, opened 2021-11-28
icons: Fix icons in non-GNOME environments

fushinari:
Fix missing icon issues in non-GNOME environments by including the missing icons ourselves.

This will also avoid theming issues to some extent at the cost of the icons possibly looking out of place in a user's theme.


fushinari:
I don't know if this is the right approach to take...

fushinari:
I don't know if this is the right approach to take...

PR #236, opened 2022-02-06
Initial libsoup async implementation

rafaelmardojai:
Changes to use libsoup async functions. Now most `Translator` methods should just process data.

## TODO

- [x] Rewrite the history code that uses `trans_queue`
- [x] Rewrite suggestions
- [x] Rewrite api key
- [x] Test and fix/implement error handing
- [ ] Optimize the code, various parts have been written just to work
- [ ] Cleanup old code

## Notes and open questions 

~With this implementation all the JSON processing code is run in the main thread, test if this doesn't affect the app fluidity. If so, maybe we can rewrite `Session.multiple()` and create other helper functions to run the processing callbacks threaded.~

~We need to figure out what to do with the TTS code. We don't have time to re-implement gTTS. Should we just leave it as is?~

~It's fine to make requests from `Translator.__init__()`?  Or should we figure out a way to make them from `DialectWindow.load_translator()` and make `Translator()` only provide the needed urls and processing functions?~

~It's `Session.multiple()` fine or should we make it run one task after another?~

### DIFF
--- a/dialect/preferences.py
+++ b/dialect/preferences.py
@@ -307,16 +307,51 @@ class DialectPreferencesWindow(Adw.PreferencesWindow):
         self.api_key.set_text(Settings.get().api_key)
 
     def _on_save_api_key(self, _button):
-        old_value = Settings.get().api_key
-        new_value = self.api_key.get_text()
+        def on_response(session, result):
+            valid = False
+            try:
+                _data = Session.get_response(session, result)
+                valid = True
+            except Exception as exc:
+                logging.warning(exc)
 
-        if new_value != old_value:
-            # Validate
-            threading.Thread(
-                target=self.__validate_new_api_key,
-                args=[new_value],
-                daemon=True
-            ).start()
+            if valid:
+                Settings.get().api_key = self.new_api_key
+                self.api_key.get_style_context().remove_class('error')
+                self.api_key_stack.set_visible_child_name('view')
+            else:
+                self.api_key.get_style_context().add_class('error')
+                error_text = _('Not a valid {backend} API key')
+                error_text = error_text.format(backend=TRANSLATORS[backend].prettyname)
+                self.error_label.set_label(error_text)
+
+            self.backend.set_sensitive(True)
+            self.backend_instance_row.set_sensitive(True)
+            self.api_key_row.set_sensitive(True)
+            self.api_key_save.set_child(self.api_key_save_image)
+            self.api_key_label.set_label(Settings.get().api_key or 'None')
+            self.instance_save_spinner.stop()
+
+        old_value = Settings.get().api_key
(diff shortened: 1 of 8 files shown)


dialect/preferences.py
```diff
@@ -209,19 +210,75 @@ def _on_edit_backend_instance(self, _button):
...
+        if self.new_instance_url != old_value:
+            # Progress feedback
+            self.backend.set_sensitive(False)
+            self.backend_instance_row.set_sensitive(False)
+            self.api_key_row.set_sensitive(False)
+            self.backend_instance_save.set_child(self.instance_save_spinner)
+            self.instance_save_spinner.start()
+
+            backend = Settings.get().active_translator
+            validation_url = TRANSLATORS[backend].format_instance_url(
```
rafaelmardojai, line 264: Is it fine or should we create `get_validation_url` and `get_settings` static methods?

fushinari: We can leave this as is for now I think. Like we discussed before, I don't think there's even going to be any other translation service with instances and/or api keys support. We don't know if the current structure will work for other services. We will probably need to re-adapt anyway.

fushinari:
> With this implementation all the JSON processing code is run in the main thread, test if this doesn't affect the app fluidity. If so, maybe we can rewrite Session.multiple() and create other helper functions to run the processing callbacks threaded.

Doesn't seem to cause any slow downs so far for me.

> We need to figure out what to do with the TTS code. We don't have time to re-implement gTTS. Should we just leave it as is?

Yeah, probably better to just leave it as is. Proxy users get broken TTS (lol) but not much we can do there. gTTS uses requests IIRC, so maybe it'll just work?

> It's fine to make requests from Translator.__init__()? Or should we figure out a way to make them from DialectWindow.load_translator() and make Translator() only provide the needed urls and processing functions?

As long as we're using the Session helper class and every request is from the same thread, I don't think it really matters much.

> It's Session.multiple() fine or should we make it run one task after another?

Should be fine.

fushinari:
> Rewrite the history code that uses trans_queue

Seems like it works just fine with a minor change.

fushinari:
So, there are some issues...
- [ ] Live Translation is somewhat broken. I can't figure out the exact reason, but it breaks on fast input or when quickly erasing text.
- [ ] The API keys validation method we implemented doesn't work as intended because LibreTranslate doesn't give an error if we use the wrong API key when API keys are optional.
- [ ] The settings UI doesnt update appropriately when instance has been reset (i.e. it doesn't hide the api key field).

There are others but I'll have to make sure, so I'll post them later.

fushinari:
That should fix all the weird re-translation issues. Hopefully, for good.

But yeah, this is very much working out fine for me. I think it might be the right direction to go.

fushinari (approved):

PR #240, opened 2022-03-03
Remove detection functions and pass "auto" lang to the service

rafaelmardojai:
Use detection capabilities through the normal translation system, this way we can detect and translate in one request.

- [x] Remove unused detection function/objects


rafaelmardojai:
We can make `get_translation` return two values, first the `Translation` object and second a detected lang code or `None`.

But the question is, it's fine the current status of always staying in "Auto"? I guess we can send a patch to LT to give the detected language in the response, but is this also possible with Google?

fushinari:
> But the question is, it's fine the current status of always staying in "Auto"? I guess we can send a patch to LT to give the detected language in the response, but is this also possible with Google?

Google already returns the detected language. But we should probably still stay in Auto to match Web UI behavior while still also somehow indicating the detected language if one is detected. The Web UI uses brackets for example.

In the future, we also need the "swap" button to be aware of the detected language, but not the language history or translation history. In the case of the translation history, I guess we can just count on the translator detecting the same language as last time.

rafaelmardojai:
> Google already returns the detected language. But we should probably still stay in Auto to match Web UI behavior while still also somehow indicating the detected language if one is detected. The Web UI uses brackets for example.

Already implemented the "Auto (Detected)" thing. I need to push it. 

> In the future, we also need the "swap" button to be aware of the detected language, but not the language history or translation history. In the case of the translation history, I guess we can just count on the translator detecting the same language as last time.

If we save the detected lang in the recen list as planned in the future, I think it should behave like that automatically

fushinari (approved):

PR #243, opened 2022-03-23
Port Google Translate back again

### LINKED ISSUE #234: [AUR Package] Error in httpcore module
I run the program from the terminal in Manjaro and gives an error:

zer0x@home:~_001

fushinari:
Honestly, I'm at a complete loss about why this doesn't work. Unless I'm missing something, it looks like it should be working but it doesn't for some reason. Google sees the request as a bad request but the equivalent code written with httpx works just fine.

Fixes #234

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,8 +39,8 @@ sudo dnf install dialect
 
 - Python 3 `python`
 - PyGObject `python-gobject`
-- GTK3 `gtk3`
-- libhandy (>= 0.90.0) `libhandy`
+- GTK4 `gtk4`
+- libadwaita (>= 1.1.0) `libadwaita`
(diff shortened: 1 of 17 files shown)


fushinari:
The changes that were needed to the existing system are basically two:

- Raw output support for `get_response` so that the bytes object is returned as is.
- Form-encoded input support for `create_message`.
- Fixed adding request headers. I'm not sure if the session user-agent takes precedence over the messsage's user-agent but I am sure that it isn't the issue here.

fushinari:
So... what's left is error handling, which is just hell in the case of gtrans...

rafaelmardojai (approved):

PR #244, opened 2022-04-03
DialectLangSelector: Fix various issues

rafaelmardojai:
- Always focus search entry
- Close language selector when Escape is pressed in search entry
- Choose the first language result when search entry is activated

### DIFF
--- a/data/resources/lang-row.ui
+++ b/data/resources/lang-row.ui
@@ -1,27 +1,18 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <interface>
-  <template class="GtkListItem">
+  <template class="DialectLangRow" parent="GtkListBoxRow">
     <property name="child">
       <object class="GtkBox">
         <property name="spacing">6</property>
         <child>
-          <object class="GtkLabel">
+          <object class="GtkLabel" id="name">
             <property name="xalign">0</property>
-            <binding name="label">
-              <lookup name="name" type="LangObject">
-                <lookup name="item">GtkListItem</lookup>
-              </lookup>
-            </binding>
           </object>
         </child>
         <child>
-          <object class="GtkImage">
+          <object class="GtkImage" id="selection">
             <property name="icon-name">object-select-symbolic</property>
-            <binding name="visible">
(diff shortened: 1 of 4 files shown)


rafaelmardojai:
I have tried to go back to `GtkListBox` and also keep the use of list models at the same time.

fushinari:
> Choose the first language result when search entry is activated

I have the following warning when I try this:

```
(dialect:2): Gtk-WARNING **: 22:51:06.681: GtkText - did not receive a focus-out event.
If you handle this event, you must return
GDK_EVENT_PROPAGATE so the default handler
gets the event as well
```

It does work though.

Additionally, would it be better or worse to bring focus onto the search bar whenever inputting text? Like we do for the main window I mean?

rafaelmardojai:
> I have the following warning when I try this:

Yeah, tried returning `Gdk.EVENT_PROPAGATE` but nothing.

> Additionally, would it be better or worse to bring focus onto the search bar whenever inputting text? Like we do for the main window I mean?

Like using `self.search.set_key_capture_widget(self)`?

fushinari:
> Yeah, tried returning `Gdk.EVENT_PROPAGATE` but nothing.

I see. I guess it's safe to ignore though?

> Like using `self.search.set_key_capture_widget(self)`?

Would that capture all events and prevent keyboard navigation? I don't really know how it behaves.

rafaelmardojai:
> Would that capture all events and prevent keyboard navigation? I don't really know how it behaves.

Yes, that property is for implementing "search as you type". This actually works, it search as you type in the popover, but the focus style of the entry is inconsistent (replacing the `grab_focus` with this).

fushinari:
> Yes, that property is for implementing "search as you type". This actually works, it search as you type in the popover, but the focus style of the entry is inconsistent (replacing the grab_focus with this).

Okay, then let's do that.

Is it possible to make it so that the user doesn't have to use Tab to go to the list? Right now, the Down arrow key does take focus away from the search entry but doesnt give it to the list. So if it's not possible, we at least don't want Down arrow key taking focus away.

fushinari (approved):

PR #247, opened 2022-05-09
Rename app id to 'app.drey.Dialect'

fushinari:
- Rename app id and resource paths to the new `app.drey.Dialect`.
- Made the changes necessary for TTS to be future proofed.
- Remove old settings and related migration code.
- Remove googletrans from flatpak and remove references to it in README.


rafaelmardojai (approved):
These changes bring so much needed oxygen in the settings side :p

Everything looks correct.

PR #250, opened 2022-05-13
window: Simplify translator object creation

fushinari:
As discussed previously, this is intended to simplify the translator object creation. But there's a problem. Trying to access Settings from the translators would cause cyclic dependency. And even when ignoring that, we can't avoid manually calling `on_loaded`. This is because it needs to be called after the `__init__()`.

About the **change default instance of LibreTranslate**. So, we keep the dance going... What do you think about just defaulting to libretranslate.com?

### DIFF
--- a/dialect/window.py
+++ b/dialect/window.py
@@ -268,14 +268,12 @@ class DialectWindow(Adw.ApplicationWindow):
         self.main_stack.set_visible_child_name('loading')
 
         # Translator object
-        if TRANSLATORS[backend].supported_features['change-instance']:
-            self.translator = TRANSLATORS[backend](
-                callback=on_loaded,
-                base_url=Settings.get().instance_url,
-                api_key=Settings.get().api_key,
-            )
-        else:
-            self.translator = TRANSLATORS[backend]()
+        self.translator = TRANSLATORS[backend](
+            callback=on_loaded,
+            base_url=Settings.get().instance_url,
+            api_key=Settings.get().api_key,
+        )
+        if not TRANSLATORS[backend].supported_features['change-instance']:
             on_loaded(True)  # Assume successful loading.
 
         if launch:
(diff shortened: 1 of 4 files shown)


dialect/window.py
```diff
@@ -268,14 +268,12 @@ def on_loaded(success, error='', network_error=False):
...
-                api_key=Settings.get().api_key,
-            )
-        else:
-            self.translator = TRANSLATORS[backend]()
+        self.translator = TRANSLATORS[backend](
+            callback=on_loaded,
+            base_url=Settings.get().instance_url,
+            api_key=Settings.get().api_key,
+        )
+        if not TRANSLATORS[backend].supported_features['change-instance']:
```
rafaelmardojai, line 276: Couldn't google just call `callback(True)`? :P

fushinari: We can't call from within __init__ because self.translator wouldn't be fully initialized yet, unlike in the case of libre or lingva.

rafaelmardojai: Hmm right, then maybe would be better if the translator object could have a `loaded` signal ig. But probably will be over complication at this point 😅.

rafaelmardojai:
> About the change default instance of LibreTranslate. So, we keep the dance going... What do you think about just defaulting to libretranslate.com?

Yeah, ig it can work. The user will have the api key needed screen right?

dialect/window.py
rafaelmardojai: > maybe would be better if the translator object could have a loaded signal

Also because I hate the callback thing xd

fushinari: > Hmm right, then maybe would be better if the translator object could have a loaded signal ig. But probably will be over complication at this point :sweat_smile:.

Hm... maybe I should try and see if it's too much... I think we already overengineered a lot of our stuff :sweat_smile: So I might end up not doing it anyway... Let's try and see ig

fushinari: Marking as WIP for this

fushinari:
> The user will have the api key needed screen right?

Yep... I'm not sure if we want that either... We have only two options really, either use libretranslate.com, which wont work without a key, making it inaccessible to many users, or keep switching between instances for each release... And I don't know which is worse. Neither option is "good" after all...

PR #261, opened 2022-06-29
Fix Flathub link

libreajans:


rafaelmardojai:
@fushinari Should we merge this and then revert it when the `2.0` version is released with the new app id? I think we changed flathub link too soon in the README.

fushinari:
Since 2.0 is coming out on July 13, I'm not sure it matters... But I guess that until then there would be a broken link on our README... Guess we can merge this for now and then change it in the release commit or later.

PR #267, opened 2022-07-18
Libadwaita 1.2

rafaelmardojai:
### Use Adw.AboutWindow
Captura desde 2022-07-17 14-22-28

### Use Adw.EntryRow

https://user-images.githubusercontent.com/6210397/179444836-9d43f5b1-cf73-421a-ba8d-8d96d241f12a.mp4

### Flat cards

Captura desde 2022-07-17 23-22-18
Captura desde 2022-07-17 23-22-29
Captura desde 2022-07-17 23-21-43

### DIFF
--- a/data/resources/about.blp
+++ b/data/resources/about.blp
@@ -1,13 +1,16 @@
 using Gtk 4.0;
+using Adw 1;
 
-AboutDialog about {
+Adw.AboutWindow about {
   modal: true;
   destroy-with-parent: true;
-  program-name: _("Dialect");
+  application-name: _("Dialect");
+  developer-name: _("The Dialect Authors");
   copyright: _("Copyright 2020–⁠2022 The Dialect Authors");
   comments: _("A translation app for GNOME.");
   website: "https://github.com/dialect-app/dialect";
-  website-label: _("GitHub page");
+  issue-url: "https://github.com/dialect-app/dialect/issues";
+  support-url: "https://github.com/dialect-app/dialect/discussions";
 
   /* Translators: Replace me with your names. */
 
--- a/data/resources/preferences.ui
+++ b/data/resources/preferences.ui
@@ -66,69 +66,33 @@
               </object>
             </child>
             <child>
-              <object class="AdwActionRow" id="backend_instance_row">
+              <object class="AdwEntryRow" id="instance_entry">
                 <property name="title" translatable="yes">Translator Instance</property>
-                <property name="subtitle" translatable="yes">Enter a translation service URL.</property>
+                <property name="tooltip-text" translatable="yes">Enter a translation service URL.</property>
+                <property name="show-apply-button">True</property>
                 <child>
-                  <object class="GtkStack" id="backend_instance_stack">
-                    <property name="transition-type">crossfade</property>
+                  <object class="GtkStack" id="instance_stack">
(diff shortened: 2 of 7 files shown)


fushinari:
Overall, I love it. Some minor issues worth mentioning though.

- The text bleeds through the scroll borders:
image
- The buttons on the bottom action bar should probably stay non-flat while the action bar itself remains flat.
- Maybe add a 4px top padding in translation-box to better space it from the headerbar, especially since it's flat now.
- Clicking reset while editing the entry should hide away the Apply button because it doesnt matter anymore.
- The API key entry's appearance and disappearance is a bit abrupt. Not really a necessary fix, but would be nice to have it appear and disappear smoothly. Revealer?

rafaelmardojai:
> The text bleeds through the scroll borders:

Yup, I think is caused because `@card_shade_color` transparency. We could use `@borders`, it's also transparent but less noticeable, but looks uglier 😅. 

> The buttons on the bottom action bar should probably stay non-flat while the action bar itself remains flat.

It looks fine for me, but not  strong opinion tbh.

> Maybe add a 4px top padding in translation-box to better space it from the headerbar, especially since it's flat now.

Right, I removed all the padding because it looked weird having more padding that in top, but i think we can work something in between.

> Clicking reset while editing the entry should hide away the Apply button because it doesnt matter anymore.

True, but not sure how to do that with `Adw.EntryRow`, maybe playing with the `show-apply-button` prop could work? Might try later.

> The API key entry's appearance and disappearance is a bit abrupt. Not really a necessary fix, but would be nice to have it appear and disappear smoothly. Revealer?

Yeah, noticed that too when implementing. But a Revealer would require us to use a ListBox inside so we can place the ActionRow right (not sure if that will break styling)? And I don't think that we want an ExpanderRow here.

fushinari:
> Yup, I think is caused because @card_shade_color transparency. We could use @borders, it's also transparent but less noticeable, but looks uglier 😅.

Oh, I see. Guess we can live with some minor bleeding. Or can we take the color and remove its transparency, maybe by mixing? I don't know what works and what doesn't in CSS.

> It looks fine for me, but not strong opinion tbh.

Yeah, me neither, but I think we want to make it consistent at least between top and bottom. And on top, it definitely looks better when not flat. Also, probably good to have contrast since it's a key functionality.

> Right, I removed all the padding because it looked weird having more padding that in top, but i think we can work something in between.

Yeah, it used to be 8 right? That would look weird. I think we can settle on 2 or 4. We do want the header to be separate but still connected.

> True, but not sure how to do that with Adw.EntryRow, maybe playing with the show-apply-button prop could work? Might try later.

Yeah, the current change should work just fine.

> Yeah, noticed that too when implementing. But a Revealer would require us to use a ListBox inside so we can place the ActionRow right (not sure if that will break styling)? And I don't think that we want an ExpanderRow here.

Yeah, that makes sense. Let's leave it as is then. It's very minor.

fushinari:
Can you also disable the switch button when the language is auto? It's broken currently.
Other than that, looks great!

fushinari (approved):

rafaelmardojai:
> Can you also disable the switch button when the language is auto? It's broken currently.

That's a long standing behavior, so let's address it on another PR.

PR #269, opened 2022-07-18
Make buttons vertical in narrow windows

### LINKED ISSUE #268: Long translations make the screen too large for mobile devices
French translations tends to be longer that English sentences. Dialect in French is then larger than in English, and does not fit anymore on a Librem5 screen.

This seems to be a common issue on GNOME apps. See [authenticator](https://gitlab.gnome.org/World/Authenticator/-/issues/341) or [Déjà Dup](https://gitlab.gnome.org/World/deja-dup/-/issues/310) for instance.

This follows the conversation in the end of #262

rafaelmardojai:
Make buttons vertical in narrow windows.

Fixes #268

Captura desde 2022-07-18 14-39-39
Captura desde 2022-07-18 14-39-46
Captura desde 2022-07-18 14-40-47
Captura desde 2022-07-18 14-42-04

### DIFF
--- a/dialect/window.py
+++ b/dialect/window.py
@@ -30,9 +30,9 @@ class DialectWindow(Adw.ApplicationWindow):
     main_stack = Gtk.Template.Child()
     error_page = Gtk.Template.Child()
     translator_box = Gtk.Template.Child()
-    retry_backend_btn = Gtk.Template.Child()
     key_page = Gtk.Template.Child()
     rmv_key_btn = Gtk.Template.Child()
+    error_api_key_btn = Gtk.Template.Child()
 
     title_stack = Gtk.Template.Child()
     langs_button_box = Gtk.Template.Child()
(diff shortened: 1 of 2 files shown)


dialect/window.py
```diff
@@ -360,9 +359,11 @@ def loading_failed(self, details='', network=False):
         self.error_page.set_title(title)
         self.error_page.set_description(description)
 
+    @Gtk.Template.Callback()
```
fushinari, line 362: Do we want to use Callback? Having the signal be connected to from another file worsens readability imo. Is there any advantage to doing it like this?

rafaelmardojai: We have duplicated buttons now, so IMO this way we avoid importing/connecting them in the code.

fushinari: I see. Then, for the sake of consistency, it might be better to use callbacks everywhere else too.

rafaelmardojai: Right, I guess we can make a followup PR for that. 

Also maybe start using `GObject.pops.prop` that is more pythonic instead of `GObject.set_prop()`.

fushinari: Yeah, that makes sense.

PR #286, opened 2022-10-06
Pin blueprint-compiler to v0.4.0 and fix an upcoming error

### LINKED ISSUE #285: Upcoming breaking change in Blueprint
Hello! Just letting you know that a breaking change in Blueprint will affect [`window.blp`](https://github.com/dialect-app/dialect/blob/main/data/resources/window.blp#L41). When [!68](https://gitlab.gnome.org/jwestman/blueprint-compiler/-/merge_requests/68) is merged, strings won't be converted to numbers anymore, and you'll need to remove the quotes on those two lines.

I've started tagging releases, and I strongly recommend pinning to one of those in the flatpak manifest and meson submodule. The latest release has the git tag `v0.4.0`. That way these changes don't break CI or contributor workflows. New releases will be announced via TWIG and the GitLab releases page.

jameswestman:
Fixes #285.

### DIFF
--- a/build-aux/flatpak/app.drey.Dialect.Devel.json
+++ b/build-aux/flatpak/app.drey.Dialect.Devel.json
@@ -25,7 +25,8 @@
                 {
                     "type": "git",
                     "url": "https://gitlab.gnome.org/jwestman/blueprint-compiler",
-                    "branch": "main"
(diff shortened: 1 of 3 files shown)


rafaelmardojai:
Thanks!

PR #288, opened 2022-10-21
Pythonic refactoring

rafaelmardojai:


rafaelmardojai:
Question, since some time ago, locally I have a file called `define.py` with these contents:

```python
"""
Fake module for linters
"""

APP_ID = ''
PROFILE = ''
RES_PATH = ''
VERSION = ''

TRANS_NUMBER = 10  # number of translations to save in history
```

So linters don't get me TOC because the `define` module doesn't exists 😅. Do you think is worth include it in our source code?

fushinari (approved):

PR #292, opened 2022-10-30
Implement new Providers model

rafaelmardojai:
## User-facing changes

- Users now can select from multiple TTS services
- Improved providers management
- Ability to edit non-active providers
- We no longer validate provider API key on preferences. (We already have a robust error view in the provider loading, and validating api keys when the provider is not initiated is more complicated)

## Main core changes

- Introduced new providers models.
- Providers subclassing improved. Better differentiation of class attrs and instance properties. Different base classes per engine.
- Providers `format_` methods now return `Soup.Message`, giving more control of the request to the provider implementation
- Providers `get_` methods now receive the data as `bytes`, giving more control on how to deal with the data
- Make custom implementations of `Gio.ListModel` to improve our data that interacts with widgets

## TODO

- [x] Update search-provider if needed `NO GOD! PLEASE NO!!! NOOOOOOOOOO`
- [x] When setting the saved selected langs, check if are present in `self.translator.languages`.
- [x] Allow having empty recent langs
- [ ] Exhaustive testing to prevent regressions before merging


rafaelmardojai:
**Current engine/feature status:**

- Local: TTS
- Soup: Translation, TTS

I think we will implement the different engines per feature when we implement some consumer. That way we don't make unnecessary work.

rafaelmardojai:
**Open questions for discussion before merging:**

- In `providers/base` I'm using python type annotation so we have some kind of documentation when implementing new providers. Should we start using it everywhere to improve code understanding for possible contributors?
- We have an inconsistent use of `@Gtk.Template.Callback()`. I'm using it on new code, but pieces of code moved in the refactoring are using the old way. If we decide to use it when possible, lets add some good comments in the callbacks definitions so it's clear from were that code is run.

@fushinari

fushinari:
image

Something is broken.

Regarding type annotation, I agree that we should be using it wherever possible. And about callbacks, wherever possible, we can have user-readable comments that still give us enough info to work with. For example: "window: Main Window (DialectApplicationWindow) -> Translate Button (translate_btn)"

rafaelmardojai:
> Something is broken.

Fixed, I was doing a really dumb thing there 😅

dialect/window.py
```diff
@@ -639,8 +648,14 @@ def _on_src_lang_changed(self, _obj, _param):
...
-            code = self.dest_langs[1] if code == self.src_langs[0] else dest_code
-            self.dest_lang_selector.selected = self.src_langs[0]
+            if len(self.dest_langs) >= 2:
+                code = self.dest_langs[1] if code == self.src_langs[0] else dest_code
+            if self.src_langs:
+                self.dest_lang_selector.selected = self.src_langs[0]
+            else:
+                options = list(self.translator.languages)
+                options.remove(code)
+                self.dest_lang_selector.selected = random.choice(options)
```
fushinari, line 658: Wait what?

rafaelmardojai: Yeah, I know 😅...

After seeing `libretranslate.de` only offering four language and weird things happening, this is a last resort thing that should set a random dest lang.

This will happen if you select a src lang that is already selected in dest, but then the provider only have one lang in recents (`libretranslate.de`  only had `en` from `['en', 'fr', 'es', 'de']`) so instead of crashing we select a random lang from `self.translator.languages` after excluding the already selected one.

So... I know it looks weird, so ideas welcome 😅.

fushinari: Yeah I understand 😅 It's fine. It's not like we really have a choice...

rafaelmardojai:
The search provider has been updated. Should be ready to merge.

@fushinari Could you test if there's any major regression? If all is OK, I think we can merge and keep testing it on master branch.

fushinari:
There was a bug in the search provider but I've fixed it now. I couldn't notice any other regression in my testing.

As for my other commit, it's a fix for the "--text" parameter not working for LibreTranslate and Lingva. I found out that this is due to the language list not  being loaded yet. Due to our asynchronous style of loading, the most reliable way for us seems to be the notify::visible-child-name signal and I've used the same.

fushinari (approved):

PR #298, opened 2022-12-05
help-overlay: Remove visible properties

libreajans:
Due to widgets are visible by default on GTK4 remove visible property names.
Source: https://gitlab.gnome.org/GNOME/gtk/-/blob/main/gtk/gtkwidget.c#L1275


rafaelmardojai:
Thanks! Looks like a leftover of the GTK4 port.

PR #309, opened 2023-02-13
Widgets: LangSelector: Sort languages model by name

### LINKED ISSUE #283: Languages list is sorted according to English name of language even when the system language is not English
The lists of languages in the list boxes are always sorted according to the language names in English.

When the system language is different (i.e. Hebrew), the names of the languages are properly translated to it, but are not re-sorted accordingly. This makes it difficult to find the desired language in the list.

The language list boxes should be sorted according to the values that they actually show.

Reported downstream [here](https://bugs.launchpad.net/ubuntu/+source/dialect/+bug/1990759).

rafaelmardojai:
Fixes #283

### DIFF
--- a/dialect/widgets/lang_selector.py
+++ b/dialect/widgets/lang_selector.py
@@ -52,8 +52,10 @@ class LangSelector(Adw.Bin):
         self.recent_model.connect('items-changed', self._on_recent_changed)
 
         self.filter = Gtk.CustomFilter()
-        self.filter.set_filter_func(self._filter_func)
-        filter_model = Gtk.FilterListModel.new(self.model, self.filter)
+        self.filter.set_filter_func(self._filter_langs)
+        sorter = Gtk.CustomSorter.new(self._sort_langs)
+        sorted_model = Gtk.SortListModel.new(model=self.model, sorter=sorter)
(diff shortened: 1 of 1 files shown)


dialect/widgets/lang_selector.py
```diff
@@ -105,10 +107,15 @@ def _popover_closed(self, _popover):
     def _create_lang_row(self, lang):
         return LangRow(lang)
 
-    def _filter_func(self, item):
+    def _filter_langs(self, item):
         search = self.search.get_text()
         return bool(re.search(search, item.name, re.IGNORECASE))
 
+    def _sort_langs(self, lang_a, lang_b, _data):
+        a = lang_a.name.lower()
```
fushinari, line 115: How does this work with non-latin languages? Any idea?

rafaelmardojai: Technically it's comparing against the numbers behind the string characters, and unicodes are ordered alphabetically so should work for all alphabets. But let me try testing it.

rafaelmardojai: Yeah, tested it and seems to be working nicely with other alphabets.

fushinari: Okay, cool!

PR #311, opened 2023-02-14
Use Python type annotations

rafaelmardojai:
Updated a few modules to use python type annotations so we can establish our criteria on how/where to use them.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,17 @@
+
+
+## Code Style
+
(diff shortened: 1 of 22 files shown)


rafaelmardojai:
@fushinari With these modules updated I think we can discuss what we want in the regard of type annotations

mufeedali:
Anything with a non-default type or variable type should be annotated I think.

mufeedali:
Also... what do you say to using something like Black?

PR #314, opened 2023-02-14
Add Microsoft provider

rafaelmardojai:
This provider is implemented to test our needs when it comes to this kind of "official" APIs with private keys. Some open questions:

- We want to have some UI to show API consumption limit?
- Provide a way to expose a short description on how to get an API key?
- Is fine to store the API key in plain GSettings?

**Blockers**:
- [ ] In some cases Microsoft requieres to pass the "Subscription-Region" string, our current API doesn't allow providers to expose this kind of "extra config fields"

### DIFF
--- a/dialect/providers/base.py
+++ b/dialect/providers/base.py
@@ -5,6 +5,7 @@
 import io
 import json
 import logging
+import urllib.parse
 
 from gi.repository import GLib, Soup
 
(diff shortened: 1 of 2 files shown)


fushinari:
> We want to have some UI to show API consumption limit?

Yes, I think showing this only in the preferences window would be fine.

> Provide a way to expose a short description on how to get an API key?

I guess it would be a nice to have. But design consideration for this is important since we don't want to crowd the UI.

> Is fine to store the API key in plain GSettings?

We need to switch to libsecret for the API keys alone I think.

mufeedali:
Since we now have bing, let's hold off on this one.

mufeedali:
Keeping the branch for reference, but I don't think we need the support itself anymore.

PR #315, opened 2023-02-14
Allow fallingback to provider lang name if we don't have it

rafaelmardojai:


rafaelmardojai:
About languages code normalization, I think we will need something more elaborated that can allow us to make comparisons between providers, like checking if al lang is available for TTS.

rafaelmardojai:
> About languages code normalization, I think we will need something more elaborated that can allow us to make comparisons between providers, like checking if a lang is available for TTS.

Committed an implementation resolving this.

But now I think the "denormalization" should be made outside right? So we don't have to call this function on every provider.

rafaelmardojai:
@fushinari Now I'm kind of happy with this, so is ready to review and merge.

fushinari (commented):
Language hinting for "Auto" seems to be broken right now. And "nonstandard" is probably more correct than "unstandard". Other than that, lgtm!

rafaelmardojai:
> Language hinting for "Auto" seems to be broken right now. And "nonstandard" is probably more correct than "unstandard". Other than that, lgtm!

Fixed both.

Tested translating from and to two languages that aren't present on `languages.LANGUAGES`:
Captura desde 2023-02-21 17-01-19

mufeedali (approved):

PR #318, opened 2023-02-23
Update Preferences window

rafaelmardojai:


rafaelmardojai:
About the changes in `window.py`, I'm still not sure if it's too much noise for validating API keys 😅

mufeedali:
Okay, this looks good in testing. I need to verify the API logic later, but would be great if you could first rebase this on the current main. Thanks for keeping things alive :)

mufeedali (approved):

PR #322, opened 2023-03-23
preferences: Add tooltip to provider settings buttons

rafaelmardojai:


rafaelmardojai:
This should be the last new strings needed before we can update translations and notify translators for the upcoming version.

PR #326, opened 2023-04-27
Update app.drey.Dialect.desktop.in.in

baarkerlounger:
Add StartupNotify to enable splash screen on Phosh while app launches


rafaelmardojai:
Thanks!

PR #328, opened 2023-05-18
Update gettext code to use two catalogues for language names and UI

rafaelmardojai:
Other complementary changes:
- Stop calling `from gettext import gettext as _` since  `_()` should be already installed on the global namespace.
- Also install `gettext()` and use it to get lang names and avoid including those strings on the regular UI po file.
- Move languages list from CLDR to `define` module as a plain dict.
- Make google provider use our lang names machinery (`BaseProvider.add_lang()`) since the `ALIASES` list is growing.

Depends on https://github.com/dialect-app/po/pull/16.


rafaelmardojai:
@mufeedali Hey, merged the new structure on `po` (https://github.com/dialect-app/po/pull/16). 

Tested this (installing the flatpak locally) and seems to be working fine, could you test it too? :p

mufeedali:
image

Hmm... Passing LANG=ar to simulate an Arabic setup, I cant see the translations for anything other than the languages.

rafaelmardojai:
> 
> Hmm... Passing LANG=ar to simulate an Arabic setup, I cant see the translations for anything other than the languages.

Discovered something funny.

if I run `LANG=es flatpak run app.drey.Dialect.Devel` I have the same issue as you. But f I just run `flatpak run app.drey.Dialect.Devel` (my system locale is spanish) I get the translations just right.

mufeedali:
Wow, that is interesting... then what would be the correct way to set per-app locale?

rafaelmardojai:
> Wow, that is interesting... then what would be the correct way to set per-app locale?

Not sure. The thing is that it only works if I use `LANG=es_PE.UTF-8`. 

But in the other hand is pretty strange that it gets the langs names right.

rafaelmardojai:
So changed my system language to Portuguese and the issue happens.

rafaelmardojai:
Tested again using the `LANGUAGE` env var and everything works fine.

mufeedali:
Okay, looks good. Thanks for confirming!

mufeedali (approved):

PR #331, opened 2023-06-26
Added image translation feature

mdpe-ir:
image

It use Gnome Screenshot And Tesseract OCR under the hood

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,11 @@ sudo dnf install dialect
 - gTTS `python-gtts`
 - D-Bus `python-dbus`
 - Beautiful Soup `python-beautifulsoup4`
+- Gnome Screenshot `gnome-screenshot`
+- Imagemagick `imagemagick`
+- Python pytesseract `python-pytesseract`
+- Tesseract Ocr `tesseract-ocr`
+- 
 
 If official packages are not available for any of the python dependencies, you can install them from pip:
 
@@ -58,6 +63,8 @@ pip install gtts
 ```
 
 ### Building from Git
+
+‍‍```
 
 ```bash
 git clone --recurse-submodules https://github.com/dialect-app/dialect.git
--- a/data/resources/window.blp
+++ b/data/resources/window.blp
@@ -404,6 +404,12 @@ template DialectWindow : Adw.ApplicationWindow {
                 }
 
                 Box {
+                  Button capture_btn {
+                    action-name: "win.capture";
+                    tooltip-text: _("Extract Text From Image");
+                    icon-name: "camera-photo";
+                  }
+
                   Button clear_btn {
                     action-name: "win.clear";
                     tooltip-text: _("Clear");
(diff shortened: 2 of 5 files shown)


mdpe-ir:
@LyesSaadi 
This will work with auto . 
But it depends on the user which package he has installed, `tesseract-ocr` only for english or `tesseract-ocr-all` for all available languages .

rafaelmardojai:
Hi, thanks for taking the time to work in this feature.

I have been asked for this functionality before and I have pointed to use [Frog](https://github.com/TenderOwl/Frog/) and paste the resulting text to Dialect. Although maybe a built-in feature would be more handy, @mufeedali  what's your opinion of about including this.

About the code, sadly I believe the correct implementation of this feature needs to be more involved. Some points to consider:

1. We can't rely on just calling `gnome-screenshot`, also it's a deprecated application so isn't present on modern GNOME installations. The screenshot functionality should use the `org.freedesktop.portal.Screenshot` portal provided by `xdg-desktop-portal`, either calling it directly with dbus or adding a dependency to `libportal`.

2. Our main distribution method is Flatpak, so we can't rely on the user installing tesseract lang packages (of course we could special-case if Flatpak isn’t detected and try to use the system packages), so we would need to add an interface that lets the user manage their OCR langs and add a flow to install them when the user tries to use the feature with a language that is not installed.

3. I think that the OCR process executed by tesseract might be heavy depending on the image, so we should run it in a separate python thread to avoid blocking the UI.

mdpe-ir:
Hi @rafaelmardojai 

Thank you for reviewing my code and providing me with your valuable feedback. I truly appreciate it. 

I looked at your frog project and it's really awesome! However, as I haven't developed a Flatpak package before, I will need some time to familiarize myself with both Flatpak and the Frog Project code so that I can apply the same knowledge to the Dialect Project In My Next Pull Request.

Regarding your suggestions for improving the image-to-text feature, I agree that they are all valid and constructive. While my aim was to showcase the potential of the built-in feature, I acknowledge that there is still a lot of work that needs to be done to enhance its

rafaelmardojai:
@mdpe-ir I'm not involved with Frog, but it's Python and it's using the technologies I mentioned, so can serve of inspiration.

mufeedali:
> Although maybe a built-in feature would be more handy, @mufeedali what's your opinion of about including this.

As a feature, I love it. But as you explained already, the implementation needs some work still.

But more importantly, do we want this as a local solution?

If we're implementing this, I would want this to be the first step towards making Dialect work fully locally. For example, we would need a models download dialog for the OCR and ideally, it would share some code with the language model downloads for any offline translation solution we build into Dialect.

Otherwise, we stick to the server-based model and implement this in LibreTranslate.

mdpe-ir:
We can add a feature in the settings section for users to obtain their API key from the website:
https://ocr.space/ocrapi
and extract text from an image online, removing the complexities of offline mode.

mufeedali:
Closing this as this implementation is not mergeable and it's too out of sync with current branch. Please let us know if you're interested in implementing this correctly, this is still a feature we're interested in :)

PR #343, opened 2023-09-24
session: Fix error leaking in multiple()

rafaelmardojai:


mufeedali (approved):

mufeedali:
Seems to be working, no crashes or anything so far. And code also looks good.

PR #345, opened 2023-09-25
search_provider: Update to intern classes changes

rafaelmardojai:


mufeedali:
Only issue I'm seeing is that Language change in the app doesn't apply to search provider unless you manually kill the search_provider process. So we should probably find a way to fix that... Any ideas?

rafaelmardojai:
> Only issue I'm seeing is that Language change in the app doesn't apply to search provider unless you manually kill the search_provider process. So we should probably find a way to fix that... Any ideas?

Yes, noticed that too. 

Problem is that currently we don´t have any signal to know that the translator changed. In the app, the prefs window calls the main window to reload the translator when the user changes the combo row. Ideally the main window should listen a signal and do it by itself.

rafaelmardojai:
I tried connecting to `Settings.get().get_child('translators')` `changed` signal, but it never notified about `active` changing.

Docs says: Note that settings only emits this signal if you have read key at least once while a signal handler was already connected for key. 

And make sense because every time we call `get_child('translators')` we're creating a new instance.

rafaelmardojai:
> And make sense because every time we call get_child('translators') we're creating a new instance.

This gave an idea

rafaelmardojai:
Ok, last commit should resolve it. I tested it in the app but didn't have time to test the SP.

I have to go out so might test it later.

mufeedali:
Just tested. Doesn't look like it fixed the search provider :(

rafaelmardojai:
Hmm, weird, it works for me.

mufeedali:
1. I search in the global search for something. "define" for example. It translates from "Auto" to the last target language (Albanian in my test).
2. Results appear.
3. Open Dialect and translate something else to another target language (French in my test).
4. Now search in Global search.
5. Still translates in Albanian.

Also, how do we get rid of the background app indicator?

rafaelmardojai:
Oh, so it doesn't update to the lasted saved recent lang. This can be fixed by connecting to the whole `self.translator.settings::changed` signal I guess, seems like I dropped that in the process. 

But remember that Dialect doesn't save recent langs until you close the app.

mufeedali:
> But remember that Dialect doesn't save recent langs until you close the app.

Yeah, but closing the main window isn't enough. You have to kill the search_provider process for the change to apply

rafaelmardojai:
> Yeah, but closing the main window isn't enough. You have to kill the search_provider process for the change to apply

Yup, I just meant that with the fix you still need to close the window 😅.

mufeedali (approved):

mufeedali:
Works as intended!

Issue #306, opened 2023-01-20
Make Font Size configurable


mchlnix:
Hello,

I have my system font set rather small, but when working with Kanjis e.g. 海岸, there is too much detail lost. Now I don't want to increase my system font size (maybe I'll have to in the future), I'd rather have the option of setting a font size just for this program.

PR #355, opened 2023-11-06
Add shortcuts to change font size

lib-retto:
This is the first step towards making the font size configurable, as requested by #332 and #306.
There are no shortcuts using the mouse wheel as it doesn't seem to be possible to add that as an accelerator.

### DIFF
--- a/data/resources/shortcuts.blp
+++ b/data/resources/shortcuts.blp
@@ -36,6 +36,16 @@ template $DialectShortcutsWindow : ShortcutsWindow {
         title: C_("shortcuts window", "Show Pronunciation");
         action-name: "app.pronunciation";
       }
+
+      ShortcutsShortcut {
+        title: C_("shortcuts window", "Increase font size");
+        action-name: "win.inc-font";
+      }
+
+      ShortcutsShortcut {
+        title: C_("shortcuts window", "Decrease font size");
+        action-name: "win.dec-font";
+      }
     }
 
     ShortcutsGroup {
--- a/dialect/main.py
+++ b/dialect/main.py
@@ -125,6 +125,8 @@ class Dialect(Adw.Application):
         self.set_accels_for_action('win.forward', ['<Alt>Right'])
         self.set_accels_for_action('win.switch', ['<Primary>S'])
         self.set_accels_for_action('win.clear', ['<Primary>D'])
+        self.set_accels_for_action('win.inc-font', ['<Primary>plus'])
+        self.set_accels_for_action('win.dec-font', ['<Primary>minus'])
         self.set_accels_for_action('win.paste', ['<Primary><Shift>V'])
         self.set_accels_for_action('win.copy', ['<Primary><Shift>C'])
         self.set_accels_for_action('win.listen-dest', ['<Primary>L'])
--- a/dialect/window.py
+++ b/dialect/window.py
@@ -124,6 +124,10 @@ class DialectWindow(Adw.ApplicationWindow):
         bus.connect('message', self.on_gst_message)
         self.player_event = threading.Event()  # An event for letting us know when Gst is done playing
 
+        # Text buffers font size
+        self.font_css_provider = Gtk.CssProvider()
+        self.font_size = 11
+
         # Setup window
         self.setup_actions()
         self.setup()
@@ -147,6 +151,14 @@ class DialectWindow(Adw.ApplicationWindow):
         clear_action.props.enabled = False
         clear_action.connect('activate', self.ui_clear)
         self.add_action(clear_action)
+
+        inc_font_action = Gio.SimpleAction.new('inc-font', None)
+        inc_font_action.connect('activate', self.ui_inc_font)
+        self.add_action(inc_font_action)
+
+        dec_font_action = Gio.SimpleAction.new('dec-font', None)
+        dec_font_action.connect('activate', self.ui_dec_font)
+        self.add_action(dec_font_action)
 
         paste_action = Gio.SimpleAction.new('paste', None)
         paste_action.connect('activate', self.ui_paste)
@@ -209,6 +221,16 @@ class DialectWindow(Adw.ApplicationWindow):
         # Listen to active providers changes
         Settings.get().connect('translator-changed', self._on_active_provider_changed, 'trans')
         Settings.get().connect('tts-changed', self._on_active_provider_changed, 'tts')
+
+        # Connect text buffers to font css provider
+        self.src_text.get_style_context().add_provider(
+            self.font_css_provider,
+            Gtk.STYLE_PROVIDER_PRIORITY_USER
+        )
+        self.dest_text.get_style_context().add_provider(
+            self.font_css_provider,
+            Gtk.STYLE_PROVIDER_PRIORITY_USER
+        )
 
     def setup_selectors(self):
         # Languages models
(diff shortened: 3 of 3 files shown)


dialect/main.py
```diff
@@ -125,6 +125,8 @@ def setup_actions(self):
         self.set_accels_for_action('win.forward', ['<Alt>Right'])
         self.set_accels_for_action('win.switch', ['<Primary>S'])
         self.set_accels_for_action('win.clear', ['<Primary>D'])
+        self.set_accels_for_action('win.inc-font', ['<Primary>plus'])
```
sabriunal, line 128: Could you also add <Primary>KP_Add

dialect/main.py
```diff
@@ -125,6 +125,8 @@ def setup_actions(self):
...
+        self.set_accels_for_action('win.dec-font', ['<Primary>minus'])
```
sabriunal, line 129: Could you also add <Primary>KP_Subtract

dialect/main.py
lib-retto: Sure

dialect/main.py
lib-retto: Sure

dialect/window.py
```diff
@@ -767,6 +789,22 @@ def ui_clear(self, _action, _param):
         self.src_buffer.props.text = ''
         self.src_buffer.emit('end-user-action')
 
+    def set_font_size(self, size):
+        self.font_size = size
+        self.font_css_provider.load_from_data(
+            "textview { font-size: "
+            + str(size)
+            + "pt; }"
+        )
```
rafaelmardojai, line 798: Let's make it a bit more compact

```suggestion
        self.font_css_provider.load_from_data(f'textview {{ font-size: { str(size) }pt; }}' )
```

dialect/window.py
```diff
@@ -124,6 +124,10 @@ def __init__(self, text, langs, **kwargs):
         bus.connect('message', self.on_gst_message)
         self.player_event = threading.Event()  # An event for letting us know when Gst is done playing
 
+        # Text buffers font size
+        self.font_css_provider = Gtk.CssProvider()
+        self.font_size = 11
```
rafaelmardojai, line 129: Should be save this to settings? What do you think @mufeedali?

dialect/main.py
```diff
@@ -125,6 +125,8 @@ def setup_actions(self):
         self.set_accels_for_action('win.forward', ['<Alt>Right'])
         self.set_accels_for_action('win.switch', ['<Primary>S'])
         self.set_accels_for_action('win.clear', ['<Primary>D'])
+        self.set_accels_for_action('win.inc-font', ['<Primary>plus', 'KP_Add'])
+        self.set_accels_for_action('win.dec-font', ['<Primary>minus', 'KP_Subtract'])
```
rafaelmardojai, line 129: We also want the modifier for the num pad keys.
```suggestion
        self.set_accels_for_action('win.inc-font', ['<Primary>plus', '<Primary>KP_Add'])
        self.set_accels_for_action('win.dec-font', ['<Primary>minus', '<Primary>KP_Subtract'])
```

rafaelmardojai (changes requested):
Thanks for taking the time! Here some comments.

dialect/window.py
mufeedali: If we're allowing it to be changed in this manner, we should definitely be saving it.

mufeedali, line 129: Wouldn't we be ignoring system defaults by forcing 11 as the default size here?

dialect/window.py
```diff
@@ -148,6 +152,14 @@ def setup_actions(self):
         clear_action.connect('activate', self.ui_clear)
         self.add_action(clear_action)
 
+        inc_font_action = Gio.SimpleAction.new('inc-font', None)
```
mufeedali, line 155: Naming here should probably indicate specifically that we're increasing or decreasing the font size. Maybe `font_size_inc` and similar? @rafaelmardojai what do you think?

data/resources/shortcuts.blp
```diff
@@ -36,6 +36,16 @@ template $DialectShortcutsWindow : ShortcutsWindow {
         title: C_("shortcuts window", "Show Pronunciation");
         action-name: "app.pronunciation";
       }
+
+      ShortcutsShortcut {
+        title: C_("shortcuts window", "Increase font size");
```
mufeedali, line 41: We should probably somehow indicate that this is only applicable to the textviews.

@rafaelmardojai any ideas?

dialect/window.py
lib-retto: Yes. According to [this](https://stackoverflow.com/questions/68492651/what-is-the-gtk4-equivalent-of-gtk-style-context-get-property) there is no direct way of getting (default) css properties in GTK4. At least I didn't find one

lib-retto: Fixed it.

dialect/window.py
rafaelmardojai: Not big deal for me, but being explicit is always good. `font-size-inc` looks good.

dialect/window.py
rafaelmardojai: @lib-retto can you implement this? We will need to add a new integer preference named `font-size` to the gsettings schema file, include helpers for that setting in `settings.py` and get that value here (for the default value on gsettings I think we can just set it to `0` and in that case the helper will return our current initial value).

In addition to that I would also add an `Adw.SpinRow` in the preference window, so the feature is more discoverable.

Feel free to implement any of those, otherwise we can do it in a following PR.

data/resources/shortcuts.blp
rafaelmardojai: Not sure about any better than `Increase text view font size` 😅. I would argue than in apps context usually font size means its content view font size, that in our case are the text views.

dialect/window.py
lib-retto: I've now implemented the setting. Should we add a new section in the preference window for this? What would that be called, "Appearance" ?

dialect/window.py
mufeedali: > Not big deal for me

Definitely not a big deal, just a nitpick XD

lib-retto: Okay, changed it

dialect/window.py
```diff
@@ -124,6 +124,10 @@ def __init__(self, text, langs, **kwargs):
         bus.connect('message', self.on_gst_message)
         self.player_event = threading.Event()  # An event for letting us know when Gst is done playing
 
+        # Text buffers font size
+        self.font_css_provider = Gtk.CssProvider()
+        self.font_size = Settings.get().font_size
```
rafaelmardojai, line 129: Maybe we can just get rid of this var and use `Settings.get().font_size` directly? Also I don't see that you're saving back the changes anywhere.

dialect/window.py
rafaelmardojai: Yes, an "Appearance" group should be fine.

dialect/window.py
lib-retto: Most GNOME programs don't remember the font size when you change it via shortcut, e.g. Text Editor. I'd say it is more natural when the size is only saved by changing the preferences.

rafaelmardojai: I see, if that is the common behaviour then let's be consistent.

dialect/window.py
lib-retto: Ok, I'm working on it

data/resources/shortcuts.blp
mufeedali: > I would argue than in apps context usually font size means its content view font size

True I guess

lib-retto:
@rafaelmardojai @mufeedali I've now added the preference to change the default font size, could you please review it?

mufeedali:
Thanks a lot @Libretto7 ! I'm really happy with this! Tested it and seems to work and code also looks good :)

HubKing:
Default Font Size doesn't work. Flatpak, 2.3.0.

Libretto7:
I assume this regression was introduced in 7504318.

rafaelmardojai:
@HubKing please report it in a new issue. But the feature works fine for me in 2.3.0 and master.


https://github.com/dialect-app/dialect/assets/6210397/cb4263ca-5380-49c0-9694-6620158d7545

@Libretto7 that commit is not present in any release.

PR #357, opened 2023-11-17
New providers API

rafaelmardojai:


dialect/window.py
```diff
@@ -1106,86 +1023,57 @@ def translation(self, _action=None, _param=None):
...
+        self.trans_dest_pron = translation.pronunciation[1]
+
+        # FIXME: Make history work again
+        # Finally, everything is saved in history
+        """self.add_history_entry(
+            original[0],
+            original[1],
+            original[2],
+            dest_text
+        )"""
```
rafaelmardojai, line 1064: For this to work again we have two options; make `Provider.translate()` have extra args that can be passed back to the `on_done` callback or make the `Translation` class capable of storing the untranslated values (every provider will have to set them, but it's trivial). 

@mufeedali opinions?

rafaelmardojai: With the second option, I guess we can make the history store `Translation` objects instead of dicts.

mufeedali: Yeah, I think extending and holding `Translation` objects makes a lot of sense. That way, we could possibly extend the history module to also use stuff like pronunciations, mistakes, etc as well.

rafaelmardojai: Done.

rafaelmardojai:
New providers API is done I think, it's really flexible and extendable in the future.

mufeedali (approved):

mufeedali:
I've committed fixes for the two issues I did find. Code looks great! As this is a pretty huge change, we may still need more testing as it stays in main. Especially maybe the search provider, which I can't test right now as I'm on Plasma.

PR #362, opened 2023-12-05
Add --selection cli option to translate primary clipboard contents

### LINKED ISSUE #347: Directly translate the selected text with translation
The text we choose can be added to the translation program with a shortcut and translated. It can work like Deepl and the gnome plugins I added below. Once we press the shortcut, the application comes to the fore and shows us the translation result. This is the result I have in mind. 
Example plugins;

https://extensions.gnome.org/extension/1756/text-translator/
https://extensions.gnome.org/extension/1849/screen-word-translate/

rafaelmardojai:
Fixes #347

This also refactors our CLI handing a bit so we don't need to have a separate route for args on launch.


rafaelmardojai:
Captura desde 2023-12-04 22-51-49

Captura desde 2023-12-04 22-54-30

dialect/main.py
```diff
@@ -36,10 +36,13 @@ def __init__(self):
...
         # App window
         self.window = None
-        self.launch_text = ''
-        self.launch_langs = {}
+        # CLI
+        self.argv = {}
+        self._signal_handler = None
 
         # Add command line options
+        self.add_main_option('selection', b'n', GLib.OptionFlags.NONE,
```
rafaelmardojai, line 44: Not really sure about the short name `n`, the `s` is already used by `src`. Also Gio docs don't say anything about the short name being nullable, not sure if a null character or something like that works for that.

mufeedali:
Need to be on GNOME to test this, so putting it off for now.

dialect/main.py
mufeedali: Ideally, no short name... but well, 'n' is fine if required

rafaelmardojai:
> Need to be on GNOME to test this, so putting it off for now. Maybe tomorrow. Or if you feel confident enough, you could merge as is too i guess.

If code looks good to you then I guess can be merged, the change is kinda small and I tested it a few times.

Btw, any limitation why you can't test it on plasma?

rafaelmardojai:
It should work on any wayland compositor implementing the [primary selection protocol](https://wayland.app/protocols/primary-selection-unstable-v1).

mufeedali (approved):
Code looks good, so let's merge. 

> Btw, any limitation why you can't test it on plasma?

No, I just wanted to test on the platform we're targeting. There are a bunch of weird issues when testing any GNOME app with a Plasma environment, usually cosmetic (cursor theme, missing icons, etc). Those don't happen when normally installing and using a GNOME flatpak though. Anyway, I'd rather rule out issues like that by default when testing something like this.

PR #363, opened 2023-12-07
Create custom TextView widget

rafaelmardojai:
Follow up of #355.

- Create a custom TextView widget to implement our needs
- Implement scroll controlled font size


dialect/window.py
```diff
@@ -234,17 +226,20 @@ def setup(self):
...
-            self.font_css_provider,
-            Gtk.STYLE_PROVIDER_PRIORITY_USER
+        # Bind text views font size
+        self.src_text.bind_property('font-size', self.dest_text, 'font-size')
+        # Set src textview mod key requirement
+        self.src_text.activate_without_mod = bool(Settings.get().translate_accel_value)
+        Settings.get().connect(
+            'changed::translate-accel',
+            lambda s, _k: self.src_text.set_property('activate_without_mod', bool(s.translate_accel_value))
         )
```
rafaelmardojai, line 236: I guess this is how you bind a int setting to a bool 😅. (Sadly we don't have access to `g_settings_bind_with_mapping` from python)

dialect/window.py
```diff
@@ -234,17 +226,20 @@ def setup(self):
...
+        if Settings.get().custom_default_font_size:
+            font_size = Settings.get().default_font_size
+            self.set_font_size(font_size)
+
+        # Set src textview mod key requirement
+        self.src_text.activate_mod = not bool(Settings.get().translate_accel_value)
+        Settings.get().connect(
+            'changed::translate-accel',
+            lambda s, _k: self.src_text.set_property('activate_mod', not bool(s.translate_accel_value))
         )
```
rafaelmardojai, line 242: I guess this is how you bind a int setting to a bool 😅. (Sadly we don't have access to `g_settings_bind_with_mapping` from python)

mufeedali: Ah well, it works and theres nothing to be done :sweat_smile:

mufeedali (approved):
There was a bug where changing the size of one view would not change the other one, I found that we werent using a bidirectional bind. I've fixed that now and I'm happy with the code and the functionality, so we should be good.

rafaelmardojai:
> There was a bug where changing the size of one view would not change the other one, I found that we werent using a bidirectional bind. I've fixed that now and I'm happy with the code and the functionality, so we should be good.

Ohh true, I totally forgot to add binding flags at all 😅

PR #369, opened 2024-01-22
Deepl API provider

rafaelmardojai:
A correct initial implementation needed dome tweaks on other parts of Dialect's code.

There are some open questions:

- ~Do we want to support pro users? DeepL has separate API domains for pro and free users (`api-free.deepl.com` and `api.deepl.com`). We could support this with our current "instances" feature, but that will require users manually setting the pro domain.~

  ~A more complex solution could be made introducing some kind of limited "provider config fields" API so providers can define some configuration fields in the form of entries, dropdows, switches, etc, this also is sort of required if we want to implement an official Microsoft provider (see  #314).~

  No plan to support `api.deepl.com` for now.

- ~Do we know the char limit? The docs mention an `128 KiB` limit, can we convert that to an approximate char limit?~
  Set to 5000 characters. 

TODO:

- [x] We need to extend the providers API to allow different src/dest lang lists. DeepL has two separated lists with some divergences.
- [x] Extend providers API to allow providers report API usage and limits. Then support the new API from the preferences window.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,11 +49,12 @@ sudo apt-get install dialect
 - PyGObject `python-gobject`
 - GTK4 `gtk4`
 - libadwaita (>= 1.4.0) `libadwaita`
+- libsoup (>= 3.0) `libsoup`
+- libsecret
 - GStreamer 1.0 `gstreamer`
 - Meson `meson`
 - Ninja `ninja`
 - gTTS `python-gtts`
-- D-Bus `python-dbus`
 - Beautiful Soup `python-beautifulsoup4`
 
 If official packages are not available for any of the python dependencies, you can install them from pip:
--- a/data/resources/provider-preferences.blp
+++ b/data/resources/provider-preferences.blp
@@ -78,6 +78,25 @@ template $ProviderPreferences : Adw.NavigationPage {
           }
         }
       }
+
+      Adw.PreferencesGroup api_usage_group {
+        title: _("Character Usage");
+        visible: false;
+
+        Box {
+          orientation: vertical;
+          spacing: 3;
+
+          LevelBar api_usage {
+
+          }
+
+          Label api_usage_label {
+
+          }
+        }
+
+      }
     }
   }
 }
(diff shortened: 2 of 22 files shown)


mufeedali:
> Do we want to support pro users?

No tbh. I don't think that's needed. 500,000 characters per month should be enough for most users. That's over 16500 characters per day. Should be enough. If not... I don't think Dialect is really meant for that anyway...

> Do we know the char limit?

Free API has a limit of 500,000 characters per month. Regarding the character limit per query, I think it's better if we enforce a limit from our end of maybe 2000 characters. 2000 characters would mean that it takes a user at least 8 queries to exceed their daily API limit (not really a limit, but the 16500 characters I mentioned earlier) and will definitely be under 128KBs as well.

> DeepL has two separated lists with some divergences.

Between Pro and Free? I'd say skip pro. I feel like DeepL is being generous enough for most users.

> Extend providers API to allow providers report API usage and limits.

Yes, we definitely need to add both UI and API support for that.

rafaelmardojai:
> No tbh. I don't think that's needed. 500,000 characters per month should be enough for most users. That's over 16500 characters per day. Should be enough. If not... I don't think Dialect is really meant for that anyway...

I guess someone could "hack" Dialect and change the `instance_url` on GSettings.

> Between Pro and Free? I'd say skip pro. I feel like DeepL is being generous enough for most users.

Nope, they have distinct langs list, for example in `src` they have English (en) and in `dest` they have both American (en_US) and British (en_GB), same for Portuguese.

I had some workaround code for that in my webkit-based implementation that I can copy, but probably would be better to rework how we handle langs to support this use case, also checking what would we need for offline providers.

mufeedali:
> I guess someone could "hack" Dialect and change the instance_url on GSettings.

My thought process is just that anyone who needs the Pro probably doesn't need Dialect at all.

> Nope, they have distinct langs list

Oh... Thats different from what I imagined. Then yeah, I guess we need a clean implementation for that.

rafaelmardojai:
> My thought process is just that anyone who needs the Pro probably doesn't need Dialect at all.

Yep, agree.

mufeedali (approved):

PR #373, opened 2024-02-26
Update search provider to translation providers API changes

rafaelmardojai:


rafaelmardojai:
@mufeedali I'm out of time these days, so I'm pushing this just in case you have time to finish it.

Technically is working from some small tests, but it might require more testing and code clean ups.

rafaelmardojai:
After some testing this seems to work fine.

@mufeedali if code looks good I can merge and release a new version.

PR #375, opened 2024-03-04
Add Debian instructions to README.md

stinerman:
There are instructions for Flathub, Arch, and Fedora. This adds instructions for Debian.

I didn't check any other Debian-based distros, but I did confirm that Debian 12 (current stable) has Dialect available for installation.


rafaelmardojai:
Thanks!

PR #378, opened 2024-03-11
Update Developer ID

yakushabb:
Appstream decided to use reverse DNS for developer IDds.

> The element should have a id property, containing a unique ID to identify the component developer / development team. It is recommended to use a reverse-DNS name, like org.gnome or io.github.ximion, or a Fediverse handle (like @user@example.org) as ID to achieve a higher chance of uniqueness.

More information: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-developer


rafaelmardojai:
Looks good, thanks!

PR #385, opened 2024-04-18
providers: deepl: Support DeepL Pro API keys

rafaelmardojai:
Not sure how `api_char_usage()` will react, since DeepL Pro is said to me unlimited.


mufeedali (approved):
I mean, we have no way to know... So let's wait for a bug report if it ever comes up... No other real choice here. LGTM otherwise :)

PR #391, opened 2024-06-07
Update the metainfo file to denote the addition of the new DeepL provider

Nekothegamer:
update the metainfo denoting the new deepl provider


rafaelmardojai:
Thanks!

PR #401, opened 2024-09-05
Experimental async

rafaelmardojai:
In this PR I'm experimenting with the recent [asyncio integration](https://gitlab.gnome.org/GNOME/pygobject/-/merge_requests/189) in PyGGbject.

On top of that I've made other refactors to make the new async code more clear and compact.

Aside from Soup's requests the only Gio async function we use is  `Gdk.Clipboard.read_text_async()`, ~~and I believe the current callback based code will be equally compact than using `create_background_task` so I'm not touching them for now~~ (With our `@background_task` decorator, that's not longer a problem). I also through about making an async version of `ProviderSettings.api_key` but I don't think is worth it for now.

After some testing everything seems pretty solid, but since it's an "experimental" feature we probably want to hold this as draft for some time.

## TODO

- [ ] Test providers APIs are working as expected 
  - [x] Translation
  - [x] TTS
  - [x] Suggestions 
  - [ ] API keys
    - [x] DeepL
    - [ ] LibreTranslate 
- [x] Port search provider

### DIFF
--- a/dialect/asyncio.py
+++ b/dialect/asyncio.py
@@ -1,0 +1,32 @@
+import asyncio
+import contextlib
+from typing import Coroutine
+
+from gi.events import GLibEventLoopPolicy
+from gi.repository import GLib
+
+
+@contextlib.contextmanager
+def glib_event_loop_policy():
+    original = asyncio.get_event_loop_policy()
+    policy = GLibEventLoopPolicy()
+    asyncio.set_event_loop_policy(policy)
+    try:
+        yield policy
+    finally:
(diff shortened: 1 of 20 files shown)


mufeedali:
So do we want to merge this right now?

rafaelmardojai:
> So do we want to merge this right now?

After we release `2.5.0` preferably. So we have longer time to keep testing it after merge.

rafaelmardojai:
@mufeedali I think we can merge this now.

PR #405, opened 2024-10-08
build: Use correct po directory for info files

bragefuglseth:
Currently, neither the metainfo file nor the desktop file are populated with translations. Since the relevant PO files are located in the /ui subdirectory of the po submodule, Meson needs to be told to look for translation files there :slightly_smiling_face:


rafaelmardojai:
Thanks! Looks like we overlooked this some time ago 😅

PR #406, opened 2024-10-09
window: Use `Adw.MultiLayoutView`

rafaelmardojai:
Use  `Adw.MultiLayoutView` for moving widgets around the desktop and mobile modes without duplicating them.

Also remove deprecated widgets like Adw.Squeezer.

### DIFF
--- a/dialect/window.blp
+++ b/dialect/window.blp
@@ -47,8 +47,10 @@ template $DialectWindow : Adw.ApplicationWindow {
   Adw.Breakpoint {
     condition ("max-width: 680px")
     setters {
-      actionbar.revealed: true;
-      title_stack.visible-child-name: "label";
+      multi_layout.layout-name: "mobile";
+      error_buttons.orientation: vertical;
+      key_buttons.orientation: vertical;
+      mobile_buttons_size.mode: horizontal;
       translator_box.orientation: vertical;
     }
   }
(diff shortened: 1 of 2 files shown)


dialect/style.css
```diff
@@ -5,7 +5,6 @@
 }
 
 .translation-box.vertical {
-  padding-bottom: 0;
```
mufeedali, line 8: This looks weird to me... I think I'd want the padding to remain the same as before :(

mufeedali (changes requested):
I'm also a bit conflicted on the new bottom button layout. I think it's a change that makes sense though. So only issue I see would be the bottom padding in vertical mode.

dialect/style.css
mufeedali: Screenshot From 2024-10-13 16-46-47

rafaelmardojai: What about `6px`? Looks more homogeneous imo.
Captura desde 2024-10-13 08-49-31

mufeedali: I still think 4 looks better, a third opinion would be nice on this :sweat_smile:

rafaelmardojai: My logic is that the toolbar has `6px`, so an extra `6px` equals the translation boxes spacing. 

> a third opinion would be nice on this 😅

I guess we should ask some designer? 😅

mufeedali: The thing is, it feels like too minor of a detail to be obsessed with, especially enough to bother a designer with 😅...

I know that your logic is sound, but 6 just "feels" off by just a tiny bit to my eyes and I can't place why

rafaelmardojai: > I know that your logic is sound, but 6 just "feels" off by just a tiny bit to my eyes and I can't place why

And I feel the same for 4px, it looks a bit smaller than all the other spacing 😅. 

I could accept 0px, since at least I can see an uniform 6px padding around the toolbar, do we agree on that?

mufeedali: Yep, zero looks good and looks uniform too.

Sorry for nitpicking over such a minor detail 😅

rafaelmardojai: No problem, I also agree that the `12 + 6px` spacing looked strange. 😅

mufeedali (approved):

PR #408, opened 2024-11-01
window: `translate_selection`: Only get clipboard when window is active

### LINKED ISSUE #407: Translate from primary selection not work
I am encountering an issue with `Dialect`, both when installed via AUR (dialect-git) and Flatpak on my Archlinux Gnome Wayland desktop environment. When running `dialect -n` from terminal and attempting to translate from the primary selection, the following error occurs:
```
Traceback (most recent call last):
  File "/app/share/dialect/dialect/window.py", line 532, in on_paste
    text = clipboard.read_text_finish(result)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
gi.repository.GLib.GError: g-io-error-quark: Cannot read from empty clipboard. (1)
``` 

Steps to Reproduce:
1. Install Dialect via AUR (dialect-git) or Flatpak.
2. In GNOME Text Editor, select a word and ensure `wl-paste --primary` can paste the content.
3. With the flatpak version, run `LANG=C flatpak run app.drey.Dialect -n`
4. `Dialect` window poped up with empty input and the above error message.

rafaelmardojai:
Fixes #407

### DIFF
--- a/dialect/window.py
+++ b/dialect/window.py
@@ -46,6 +46,7 @@ class DialectWindow(Adw.ApplicationWindow):
 
     # Properties
     translator_loading: bool = GObject.Property(type=bool, default=True)  # type: ignore
+    selection_translation_queued: bool = GObject.Property(type=bool, default=False)  # type: ignore
 
     # Child widgets
     menu_btn: Gtk.MenuButton = Gtk.Template.Child()  # type: ignore
@@ -105,6 +106,7 @@ class DialectWindow(Adw.ApplicationWindow):
     current_history = 0  # for history management
 
     # Translation-related variables
+    selection_translation_langs: tuple[str | None, str | None] = (None, None)
     next_translation: TranslationRequest | None = None  # for ongoing translation
     translation_loading = False  # for ongoing translation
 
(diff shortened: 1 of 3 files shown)


zefr0x:
I'm not sure about the UX, should we also have an visual indicator?

rafaelmardojai:
> The fix itself works. But wouldn't it make more sense to split the two modes of operation of `translate_selection()` into two functions? Something like `queue_selection_translation` and `translate_selection` maybe?

Done, code should be cleaner.

I also made `selection_translation_queued` a `GObject` property so it can be binded to UI if we implement what @zefr0x suggests.

dialect/window.py
```diff
@@ -511,8 +513,20 @@ async def translate_selection(self, src_lang: str | None, dest_lang: str | None)
...
+                if text := await clipboard.read_text_async():  # type: ignore
+                    self.translate(text, src_lang, dest_lang)
+            except GLib.Error as exc:
+                logging.error(exc)
+                self.send_notification(_("Couldn’t read selection clip board!"))
+
+    def queue_selection_translation(self, src_lang: str | None, dest_lang: str | None):
+        """Call `translate_selection` or queue it until the window is focused"""
+        if self.props.is_active:
+            self.translate_selection(src_lang, dest_lang)
```
mufeedali, line 526: Is this needed? Wouldn't `queue_selection_translation` only be called before the active event anyway? Or am I wrong about how the events work?

rafaelmardojai: No, `queue_selection_translation` is called by `process_command_line` in the app and does the queue only if the window is not active.

mufeedali: If I'm understanding correctly, `process_command_line` is always only called once for each instance of Dialect and that would be before the `is-active` event. So, we don't really need the`is-active` check and `translate_selection` call, right?

rafaelmardojai: `process_command_line` is called every time the app is activated by CLI, we don't support multiple Dialect instances, so it will be always called in the main one every time the user runs the command.

rafaelmardojai: >  So, we don't really need theis-active check and translate_selection call, right?
Remember that initially I could not reproduce the issue because `translate_selection` was called with the window already focused.

rafaelmardojai: I only  could reproduce the issue by activating the cli with an already opened Dialect window.

mufeedali: > `process_command_line` is called every time the app is activated by CLI...

Oh yeah, that makes sense :sweat_smile:

mufeedali (approved):

PR #410, opened 2024-11-01
docs: Replace AUR reference for stable release with Extra repo ref

mufeedali:
Dialect is now available from Arch's official repositories :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,9 +24,9 @@ A translation app for GNOME.
 
 <a href='https://flathub.org/apps/details/app.drey.Dialect'><img alt='Download on Flathub' src='https://flathub.org/api/badge?svg&locale=en'/></a>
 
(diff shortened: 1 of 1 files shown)


mufeedali:
I forgot to ask the big question here. Do we want to indicate that Flatpak is our own official package while the others are technically from the respective packagers? I don't think it's really needed anymore, but what do you think, @rafaelmardojai ?

rafaelmardojai:
> I forgot to ask the big question here. Do we want to indicate that Flatpak is our own official package while the others are technically from the respective packagers? I don't think it's really needed anymore, but what do you think, @rafaelmardojai ?

I guess it could add clarity in who maintains what, but also don't think is really needed.

rafaelmardojai (approved):

PR #425, opened 2025-02-19
refactor(asyncio): Update to use new PyGObject APIs and recomendations

rafaelmardojai:


mufeedali (approved):
This is just so much nicer... LGTM and works from my testing!

PR #426, opened 2025-02-28
language

johnpetersa19:
The Portuguese language is missing in translation providers

### DIFF
--- a/dialect/providers/modules/google.py
+++ b/dialect/providers/modules/google.py
@@ -196,6 +196,7 @@ TRANSLATE_TLDS = (
     ".pn",
     ".ps",
     ".pt",
+    ".pt_BR",
     ".ro",
     ".rs",
     ".ru",
--- a/dialect/providers/modules/yandex.py
+++ b/dialect/providers/modules/yandex.py
@@ -109,6 +109,7 @@ class Provider(SoupProvider):
             "fa",
             "pl",
             "pt",
+            "pt_BR",
             "pa",
             "ro",
             "ru",


dialect/providers/modules/google.py
```diff
@@ -196,6 +196,7 @@
     ".pn",
     ".ps",
     ".pt",
+    ".pt_BR",
```
rafaelmardojai, line 199: This is a list of Top-Level Domains, not languages. The Google provider already will fetch the languages list from Google's website (see  #411), but that hasn't been released yet.

dialect/providers/modules/yandex.py
```diff
@@ -109,6 +109,7 @@ async def init_trans(self):
             "fa",
             "pl",
             "pt",
+            "pt_BR",
```
rafaelmardojai, line 112: Have you tested this? I see that Yandex web UI only shows Brazilian Portuguese as a destination language, so maybe that could break something, since this will add the language in both source and destination lists. I will try to test whenever I get some free time.

If it needs fixing you can check the DeepL provider on how to manage divergent source and destination languages lists.

rafaelmardojai (changes requested):

dialect/providers/modules/yandex.py
johnpetersa19: Yandex translates into European Portuguese. The translation is not the same as Brazilian Portuguese. Only Bing offers the option for European Portuguese. Providers do not offer the option to choose between Brazilian Portuguese and European Portuguese.

johnpetersa19: Captura de tela de 2025-02-28 16-28-30

johnpetersa19: Captura de tela de 2025-02-28 16-29-18

johnpetersa19: The images below show the modification in Google. Yandex did not break the project and now displays the options "Brazilian Portuguese" and "European Portuguese (Portugal)

Captura de tela de 2025-02-28 16-46-29
Captura de tela de 2025-02-28 16-46-49

rafaelmardojai: In Google, as I told you in the [other comment](https://github.com/dialect-app/dialect/pull/426#discussion_r1975690691), that works since this PR (#411) that hasn't been released yet, so if you remove your change it will still work in the `master` version. Also in your change you're adding `.pt_BR` to the TLDs list, so when we pick a random Google domain you're adding `translate.google.pt_BR` to the options, which is an invalid domain.

About Yandex, what I mean is that they just list "Portuguese" as a source language, and in you change you'll get both "Brazilian Portuguese" and "Portuguese" in both lists. So you say that `Auto -> Brazilian Portuguese` works fine, what I want you to test is if  `Brazilian Portuguese -> English (or any other lang)` also works.

Also the selector will allow you to select `Brazilian Portuguese -> Portuguese` in Yandex, test if that doesn't breaks anything.

In cases like Yandex with divergent source and destination lists, as I told you, ideally you need to do what we do for DeepL.

johnpetersa19: I already modified this in the Fly project, but I don't remember how I made this correction. However, it is giving error in the request, as you mentioned. Hahaha.

PR #428, opened 2025-03-18
new function option

johnpetersa19:
Opções de aplicativo:
  -n, --selection            Translate text from the primary clipboard
  -t, --text                 Text to translate
  -s, --src                  Source lang code
  -d, --dest                 Destination lang code
  -c, --copy                 Copy translated text to clipboard

### DIFF
--- a/dialect/main.py
+++ b/dialect/main.py
@@ -114,7 +114,7 @@ class Dialect(Adw.Application):
         text = ""
         langs: dict[str, str | None] = {"src": None, "dest": None}
         selection = "selection" in self.argv
-        copy_to_clipboard = "copy" in self.argv  # Verificando a opção --copy
+        copy_to_clipboard = "copy" in self.argv  # Checking the --copy option
 
         if "text" in self.argv:
             text = self.argv["text"]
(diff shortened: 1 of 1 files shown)


mufeedali:
Hi @johnpetersa19 ,

Have you tested this?

mufeedali:
Can I also ask what the use case for this is?

johnpetersa19:
Hi @mufeedali,


I am unable to compile because an error occurs in Blueprint. When compiling, an error occurs that is not related to the changes. I am trying to test it. I apologize, but I am unable to test it before showing this possible new option.

The use of this option is to allow the translation of a text directly from the clipboard or from a manually provided text, specifying the source and target languages. In addition, there is the option to copy the automatically translated text to the clipboard.

This can be useful for automating quick translations in scripts, integrating with other tools or simply to facilitate translation without the need to manually open a translator.

johnpetersa19:
error I can't resolve to test the project


Captura de tela de 2025-03-19 22-00-41

mufeedali:
> This can be useful for automating quick translations in scripts, integrating with other tools or simply to facilitate translation without the need to manually open a translator.

Dialect is not suited to CLI or script use cases. There are plenty of tools out there that would serve your use case better than Dialect could. Even if your changes worked, Dialect wouldn't exit after the copy, for example, and we wouldn't want Dialect, a GUI-focused application, to exit there either.

Regarding the build process, the easiest way to try and build Dialect is to setup [flatpak](https://flatpak.org/setup/) on your system and run the project with [GNOME Builder](https://flathub.org/apps/org.gnome.Builder).

I will be closing this PR as I think we don't want this feature within Dialect. The existing CLI flags make sense for purposes like global shortcuts and augmenting and enhancing the GUI experience. But I don't think this new flag satisfies either of those criteria.

PR #431, opened 2025-04-13
Add D-Bus service file

City-busz:
And mark the application as D-Bus activatable. This allows application launchers to activate it via D-Bus.

Reference:
https://specifications.freedesktop.org/desktop-entry-spec/1.3/dbus.html

### DIFF
--- a/data/app.drey.Dialect.desktop.in.in
+++ b/data/app.drey.Dialect.desktop.in.in
@@ -4,6 +4,7 @@ Version=1.0
 Type=Application
 Terminal=false
 StartupNotify=true
+DBusActivatable=true
 Exec=dialect
 Name=Dialect
 Comment=Translate between languages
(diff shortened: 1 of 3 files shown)


City-busz:
This is needed for better integration with GNOME. See: https://developer.gnome.org/documentation/guidelines/maintainer/integrating.html#d-bus-activation

When `DBusActivatable=true` is set in the desktop file, the application launcher of GNOME Shell and KDE Plasma launches the application via D-Bus instead of `fork()`/`exec()` the executable, so the D-Bus daemon will control the process by setting up its environment, adding it to the correct cgroup, just like how usual D-Bus services work. And when the application is already running, the primary instance will receive the activation or the desired action directly.
If the service file is installed, the app can be launched via D-Bus with the following command:
```
gapplication launch app.drey.Dialect
```

PR #434, opened 2025-05-20
feat(providers/soup): Implement retries with exponential backoff

### LINKED ISSUE #433: DeepL provider fails on startup
As reported in https://github.com/dialect-app/dialect/issues/412#issuecomment-2892026717 DeepL provider is failing on init.

After some debugging, we're hitting a `HTTP 429: too many requests` error. Probably related to the fact that we're doing three request (src langs, dest langs and API key validation) one after the other.

rafaelmardojai:
This is needed by the DeepL provider, and they recommend it in their API docs, since the service is expected to return "HTTP 429: too many requests" errors when sending many API requests in a short period of time.

https://developers.deepl.com/docs/best-practices/error-handling

Fixes #433

### DIFF
--- a/dialect/providers/modules/deepl.py
+++ b/dialect/providers/modules/deepl.py
@@ -35,6 +35,7 @@ class Provider(SoupProvider):
     def __init__(self, **kwargs):
         super().__init__(**kwargs)
 
+        self.retry_errors = (429,)
         self.chars_limit = 5000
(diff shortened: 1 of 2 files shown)


mufeedali:
LGTM. I'm still unable to reproduce the issue from my end though... Very odd.

PR #442, opened 2025-08-26
feat: Add Kagi as a provider

mufeedali:
Session token can be taken from Kagi settings and used for testing. Trial token should also work.

Most importantly, with this, we now support "Gen Z" as a language in Dialect.

Screenshot From 2025-08-26 19-10-17

### DIFF
--- a/dialect/providers/modules/deepl.py
+++ b/dialect/providers/modules/deepl.py
@@ -85,8 +85,6 @@ class Provider(SoupProvider):
             return True
         except (APIKeyInvalid, APIKeyRequired):
             return False
-        except Exception:
-            raise
 
     async def translate(self, request):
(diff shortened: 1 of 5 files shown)


dialect/providers/modules/deepl.py
```diff
@@ -1,5 +1,5 @@
-# Copyright 2024 Mufeed Ali
```
rafaelmardojai, line 1: This should stay with the year when was first introduced.

mufeedali:
I was attempting to validate the session token by fetching the language list but the language list fetch seems to work regardless of the session token being incorrect. So I'm looking for another way to validate the API key without causing actual compute on Kagi's end. Moving to draft state for now.

rafaelmardojai (approved):

PR #445, opened 2025-09-28
chore: Remove Lingva

mufeedali:
As discussed in the past, Lingva has been unmaintained for 3 years now. This PR will be removing Lingva. More importantly, we shouldn't be recommending it and we want to remove it from our self-hosted services in the future as well. Hopefully, we will be able to implement other services to make up for it.


rafaelmardojai:
Looks good to me. 

Do we want to add SimplyTranslate support in the same release? It is not super maintained, but at least has more recent commits and is still working. Also in any case, the codebase should be easier to understand and patch for us in case we need.

